PR #77, opened 2015-02-04
Simplify date object processing

graingert:
```python
>>> issubclass(datetime.datetime, datetime.date)
True
```


spulec:
Thank you!

Issue #75, opened 2015-01-23
tests/utils.py seems to be missing from the 0.2.8 tarball


keszybz:
Tests fail with
```
ImportError: cannot import name 'utils'
```

spulec:
I had not intended for the tests to be included with the distributions. That is now fixed with d8df201256de7013b04ea57d8c4f1878787b0fe8.

Thanks.

keszybz:
Hm, but why? Running tests is a nice way to check that things are OK. Not just in freezegun itself, but e.g. in dependencies.

spulec:
[This](https://github.com/spulec/freezegun/issues/27) was the original reason. I'm open to alternative suggestions, but tend to be -0 on things like a nested tests directory inside of the freezegun module.

pypingou:
You can ship the tests in the tarball without nesting them in the module itself (because they should not be installed) which is what you do now, thanks for that! :)

keszybz:
I think that #27 was bogus. If I revert HEAD (so I'm effectively at freezegun@7da4cef779),
```
pip install .
```
does not install tests. Neither does
```
pip install freezegun==0.2.8
```
Neither does
```
python setup.py install --user
```
Nor
```
python setup.py sdist
pip install `pwd`/dist/freezegun-0.2.8.tar.gz --user  
```
I think that the reporter from #27 might have had an old egg in his git checkout directory.

Issue #27, opened 2013-12-19
freezegun installs a `tests` module.


lukesneeringer:
When installing `freezegun` with `pip install freezegun`, there seems to be a side effect, which is that it installs its tests into site-packages as well.

This can have a negative effect for tests in a `tests/` directory that need to import from other modules under said directory.

I assert that `packages=find_packages(),` in setup.py should change to:

``` python
    provides=[
        'freezegun',
    ],
    packages=[i for i in find_packages() if i.startswith('freezegun')],
```

PR #80, opened 2015-02-08
Include the tests in the releases but do not install them

pypingou:
Updating the MANIFEST.in file allows making sure the tests are included in the release tarball generated via ``python setup.py sdist`` but the ``setup.py`` needed a small adjustment to prevent it from installing them.

Fixes https://github.com/spulec/freezegun/issues/75
Fixes https://github.com/spulec/freezegun/issues/27


spulec:
Thanks. This definitely seems like the right way to do it.

PR #83, opened 2015-02-10
Remove should; make pypy tests pass

msabramo:
It looks like `should` might not work on pypy.


spulec:
I knew this day would eventually come....

PR #86, opened 2015-03-05
class decoration must respect staticmethods and other non-test callables

jesteria:
In upgrading from 0.1.18, my app's test suite ran into a heap of trouble with freezegun's class decorator interface. I still don't think I've tracked down all of it &ndash; and I've had to undo the app's use of the class decorator for now &ndash; but the most visible issue stemmed from the changes to `_freeze_time.__call__` and `decorate_class`, which were previously much more strict about which entities to patch.

I don't propose to return to a `unittest`-centric decorator, but I think these changes are reasonable; so long as we're skipping `classmethod`, we really don't want to patch `staticmethod` and non-method callables. As a concrete example, we have tests like the following:

    import mock
    from unittest import TestCase

    class TestThings(TestCase):

        urllib2_patch = mock.patch('urllib2.urlopen', …)

        @staticmethod
        def do_some_help(some_input):
            …

        @urllib2_patch
        def test_1(self, _urllib2_mock):
            helpful_output = self.do_some_help(some_input)
            …

Without these changes, the module fails on import, as freezegun attempts to wrap the `mock` library's patch object &ndash; but raises `AttributeError` on `__name__`. Or, `test_1` fails, because freezegun has wrapped `do_some_help`, such that it's no longer a `staticmethod` and takes additional arguments ("self").

Anyway, thanks for the useful library!

### DIFF
--- a/freezegun/api.py
+++ b/freezegun/api.py
@@ -235,11 +235,11 @@ class _freeze_time(object):
                 continue
 
             attr_value = getattr(klass, attr)
-            if not hasattr(attr_value, "__call__"):
-                continue
 
-            # Check if this is a classmethod. If so, skip patching
-            if inspect.ismethod(attr_value) and attr_value.__self__ is klass:
+            # Skip patching classmethods, staticmethods and non-method callables
+            # (but do recurse into nested classes)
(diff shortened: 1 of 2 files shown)


spulec:
This is great! Thank you!

PR #88, opened 2015-03-07
class decoration must ignore nested classes

jesteria:
I figured out what else was causing tests to fail &ndash; the class decorator's wrapping arbitrary methods on classes nested under the test class.

The Django `TestCase` has a reference to the HTTP request `Client` class &ndash; defined on `SimpleTestCase.client_class`. (This allows you to override the class that's initialized and set on the test case instance at `self.client`.) Because `_freeze_time.decorate_class` wasn't excluding nested classes, it would recurse into such classes and patch their methods, including `Client.get`, `.post`, *etc.*.

This might not be an issue for the tests in the decorated test class, but any subsequent tests that didn't *re-patch* this nested class's methods (by decorating themselves with `freeze_time`) would see crazy side-effects &ndash; for example, their sessions would be created far in the past, such that they were immediately expired, and so session data would never set. Tests which didn't even import freezegun would pass when run alone but fail when run as part of the full suite.

I agree that freezegun shouldn't assume a `unittest.TestCase` class interface, but I'm not sure why it *should* assume that the methods of nested classes are valid targets; and, as in the example above, to do so has pernicious side-effects.

For that matter, I think it would be nice to further restrict the decorator from touching non-function callables (such as the "mock" object). And, there's really no need for it to touch any methods but those which are actually tests; but, we don't know what the test framework / test run will consider a test, and I don't see a lot of harm in continuing to patch every function.

This change slightly refactors the work of the last pull and makes decoration exclude nested classes. (To see the full change relative to when I started, see [here](https://github.com/spulec/freezegun/compare/0.3.1...jesteria:even-stricter-class-decoration?expand=1#diff-80ef255f3bb85cfd652668d719317c10R232).)

I think it still might be a good idea to further exclude callables which don't satisfy `inspect.isfunction()` &ndash; a test collection defined by placing non-function callable objects under a class definition is a pretty weird one; and, `isfunction()` would exclude "mock" and "patch" objects. But, with the last pull, it isn't a problem for me to continue to patch these.

### DIFF
--- a/freezegun/api.py
+++ b/freezegun/api.py
@@ -230,35 +230,21 @@ class _freeze_time(object):
         return self.decorate_callable(func)
 
     def decorate_class(self, klass):
-        klass_mro = klass.mro()
-        for attr in dir(klass):
-            if attr.startswith("_"):
-                continue
+        seen = set()
+        for base_klass in klass.mro():
+            for (attr, attr_value) in base_klass.__dict__.items():
+                if attr.startswith('_') or attr in seen:
+                    continue
+                seen.add(attr)
 
-            attr_value = getattr(klass, attr)
-            if not callable(attr_value):
-                continue
+                if not callable(attr_value) or inspect.isclass(attr_value):
+                    continue
 
-            for base_klass in klass_mro:
                 try:
-                    attr_descriptor = base_klass.__dict__[attr]
-                except KeyError:
-                    pass
-                else:
-                    break
-            else:
-                # don't sweat the attribute magic
-                attr_descriptor = None
-
-            # Skip patching classmethods and staticmethods
-            if isinstance(attr_descriptor, (classmethod, staticmethod)):
-                continue
-
-            try:
-                setattr(klass, attr, self(attr_value))
-            except (AttributeError, TypeError):
-                # Sometimes we can't set this for built-in types and custom callables
-                continue
+                    setattr(klass, attr, self(attr_value))
+                except (AttributeError, TypeError):
+                    # Sometimes we can't set this for built-in types and custom callables
+                    continue
         return klass
 
     def __enter__(self):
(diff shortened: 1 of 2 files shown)


freezegun/api.py
```diff
@@ -230,35 +230,21 @@ def __call__(self, func):
...
-                except KeyError:
-                    pass
-                else:
-                    break
-            else:
-                # don't sweat the attribute magic
-                attr_descriptor = None
-
-            # Skip patching classmethods and staticmethods
-            if isinstance(attr_descriptor, (classmethod, staticmethod)):
```
spulec: I'm a bit confused why we are removing this part? Am I missing that this would now be redundant due to something else?

spulec:
You make good points and I agree, especially if this is necessary to get the Django `TestCase` working.

freezegun/api.py
jesteria: Right. All I've done in this revision is to refactor it such that, rather than get the class's internal namespace more than once, (once with `dir()`, and then iterating over the MRO for each name), it relies on only one loop over the MRO, with a `set` to keep track of names we've already seen (because they were overwritten in the subclass). This is basically what `dir()` was doing for us in the first place, (excluding classes with exotic metaclasses, which would cause problems anyway).

I started looking through the MRO, and each of its classes' `__dict__`s, in the first place, to avoid having to guess which class attribute values are proper functions (which act as instance methods), and which are instances of `classmethod` and `staticmethod`. (For that matter, this should now exclude properties as well, which makes sense.) Because rather, these are implemented in the class `__dict__` as non-callable descriptors, (which wrap functions) &ndash; and so the `not callable(attr_value)` [line](https://github.com/spulec/freezegun/pull/88/files#diff-80ef255f3bb85cfd652668d719317c10R240) can now exclude these without further inspection. So, freezegun will only wrap undecorated functions, (and callable objects, unless we start checking with `inspect.isfunction`).

spulec: Got it. That makes sense.

spulec:
Thank you again!

jesteria:
:fireworks: :dancers: :fireworks:

PR #90, opened 2015-03-20
Fix FakeTime.__call__() to return the correct value when daylight savings is in effect.

tleach:
The previous implementation was dependent on `mktime` which assumes the provided timetuple is in _local_ time.
The `time_to_freeze` is UTC, so `__call__` was attempting to compensate for this by using `time.timezone` to offset `time_to_freeze` first.
The problem is that `time.timezone` ignores daylight savings. When daylight savings is in effect, `time.timezone` is wrong by an hour and consequently `__call__` returns a timestamp which is an hour earlier than it should be.

This issue can be addressed, and the code made much cleaner, by using simply `calendar.timegm` instead of `mktime`. `calendar.timegm` assumes the given time is in UTC so no timezone adjustment is needed and timestamp will always be correct, daylight savings or otherwise.

### DIFF
--- a/freezegun/api.py
+++ b/freezegun/api.py
@@ -3,6 +3,7 @@ import functools
 import inspect
 import sys
 import time
+import calendar
 
 from dateutil import parser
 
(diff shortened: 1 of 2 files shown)


spulec:
Great, thank you!

tleach:
Thanks for merging. When are you next planning to cut a release?

spulec:
`0.3.2` is now released.

PR #91, opened 2015-03-25
Restore previous time.gmtime function in stop()

bobbyi:
The _freeze_time.stop() method needs to restore time.gmtime() to its previous value as it does for time.time().


spulec:
Thank you!

PR #92, opened 2015-04-13
Improves handling of date and datetime objects in freeze_time

daveFNbuck:
By converting date and datetime objects to isoformat and then using dateutil to
convert to datetime, this freeze_time becomes exposed to more bugs in dateutil
than strictly necessary. I have found when using hypothesis-datetime to test my
own libraries in combination with freeze_gun that years earlier than 100 cause
bugs due to dateutil interpreting them as being in the 20th or 21st centuries.

This bug is avoided by simply preserving the date or datetime rather than
serializing and deserializing it.

### DIFF
--- a/freezegun/api.py
+++ b/freezegun/api.py
@@ -217,7 +217,12 @@ def pickle_fake_datetime(datetime_):
 class _freeze_time(object):
 
     def __init__(self, time_to_freeze_str, tz_offset, ignore):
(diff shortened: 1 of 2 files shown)


spulec:
Thank you!

daveFNbuck:
No problem, just glad to have the bug patched :)

PR #104, opened 2015-07-23
Improve memory usage: forget about undo_changes after they are undone

adamchainz:
List of 1000s of 3-tuples * many uses in tests = high memory use  :(

Also repeat use of the same `_freeze_time` object would continue adding to the existing list and then do multiple times the work necessary on `stop()`.


spulec:
Thanks!

PR #108, opened 2015-08-05
Add tick to keep time moving

zmsmith:
I know this can't be merged, but this is my current overall thinking on approach. I'd probably move the factory classes into a strategies file or something and then mimic a greater subset or tests. Also happy to name the variable to time_to_freeze_strategy or something.

Let me know what you think

### DIFF
--- a/freezegun/api.py
+++ b/freezegun/api.py
@@ -5,6 +5,7 @@ import sys
 import time
 import calendar
 import unittest
+import platform
 
 from dateutil import parser
 
@@ -20,10 +21,15 @@ try:
 except ImportError:
     import copyreg
 
+
 # Stolen from six
 def with_metaclass(meta, *bases):
     """Create a base class with a metaclass."""
     return meta("NewBase", bases, {})
+
+
+def _is_cpython():
+    return platform.python_implementation() == "CPython"
 
 
 class FakeTime(object):
@@ -33,7 +39,8 @@ class FakeTime(object):
         self.previous_time_function = previous_time_function
 
     def __call__(self):
-        return calendar.timegm(self.time_to_freeze.timetuple()) + self.time_to_freeze.microsecond / 1000000.0
+        current_time = self.time_to_freeze()
+        return calendar.timegm(current_time.timetuple()) + current_time.microsecond / 1000000.0
 
 
 class FakeLocalTime(object):
@@ -43,7 +50,7 @@ class FakeLocalTime(object):
     def __call__(self, t=None):
         if t is not None:
             return real_localtime(t)
-        shifted_time = self.time_to_freeze - datetime.timedelta(seconds=time.timezone)
+        shifted_time = self.time_to_freeze() - datetime.timedelta(seconds=time.timezone)
         return shifted_time.timetuple()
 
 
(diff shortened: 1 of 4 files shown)


spulec:
Looks great. `tick` is perfect.

zmsmith:
I've pushed an updated and full implementation. Let me know if it needs anything.

adamchainz:
This is great, I've wanted this since it stops the weird world where every instruction executes at the same time. However `tick` should affect everything, not just `datetime.datetime.now()`... so `time.time()`, `datetime.date.today()` etc. too.

zmsmith:
I added an explicit test for time (which was missing prior, though there was a test for date). Let me know if I'm missing any cases:

https://github.com/zmsmith/freezegun/blob/93b532e57b5aa6466c65b26d545ce1e4b1e15baa/tests/test_ticking.py

spulec:
Thanks!

tests/utils.py
```diff
@@ -7,3 +11,12 @@ def is_fake_date(obj):
 
 def is_fake_datetime(obj):
     return obj.__class__ is FakeDatetime
+
+
+def cpython_only(func):
```
adamchainz: make `is_cpython` a variable rather than a function, then you can just do `cpython_only = unittest.skipUnless(is_cpython, "This functionality only works on CPython")` rather than making a whole decorator.

freezegun/api.py
```diff
@@ -20,12 +21,17 @@
...
 
+
 # Stolen from six
 def with_metaclass(meta, *bases):
     """Create a base class with a metaclass."""
     return meta("NewBase", bases, {})
 
 
+def _is_cpython():
+    return platform.python_implementation() == "CPython"
```
adamchainz: On the old `ctypes` branch, @spulec has the cpython test checking for whether `platform` has even the `python_implementation` attribute - see https://github.com/spulec/freezegun/commit/f429b5ae015a11c2cd25feb2ce06f868aad226de#diff-aa9b211ef57daf3482592c4cc9180558R23 . I don't know in what python interpreters this is the case, but I guess best to act defensively as I don't think this would just be added randomly.

tests/test_ticking.py
```diff
@@ -0,0 +1,42 @@
...
+import mock
+
+from freezegun import freeze_time
+from tests import utils
+
+
+@utils.cpython_only
+def test_ticking_datetime():
+    with freeze_time("Jan 14th, 2012", tick=True):
+        assert datetime.datetime.now() > datetime.datetime(2012, 1, 14)
```
adamchainz: insert a tiny `sleep()` here of `0.1` seconds, or even `0.01` seconds - some system clocks don't have microsecond resolution which means the test could intermittently fail there.

tests/test_ticking.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+@utils.cpython_only
+def test_ticking_date():
+    with freeze_time("Jan 14th, 2012, 23:59:59.9999999", tick=True):
+        assert datetime.date.today() == datetime.date(2012, 1, 15)
```
adamchainz: again, a tiny delay :)

tests/test_ticking.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+@utils.cpython_only
+def test_ticking_time():
+    with freeze_time("Jan 14th, 2012, 23:59:59", tick=True):
+        assert time.time() > 1326585599.0
```
adamchainz: ditto

tests/test_ticking.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+@mock.patch('freezegun.api._is_cpython', lambda: False)
+def test_pypy_compat():
```
adamchainz: test_cpython_compat ?

tests/test_ticking.py
```diff
@@ -0,0 +1,42 @@
...
+    try:
+        freeze_time("Jan 14th, 2012, 23:59:59", tick=True)
+    except SystemError:
+        pass
+    else:
+        raise AssertionError("tick=True should not error on cpython")
```
adamchainz: Maybe more readable with `nose`'s `raises`: http://nose.readthedocs.org/en/latest/testing_tools.html#nose.tools.raises

adamchainz:
It's okay I'll split my review feedback into a PR

PR #109, opened 2015-08-06
Tidy up tick=True

adamchainz:
Also make tox install requirements.txt and add @adamchainz to AUTHORS.


adamchainz:
Basically some stuff I commented on #108. I was mistaken about unittest `skipUnless` since ofc Python 2.6 doesn't include it :scream_cat:

spulec:
Great, thanks!

PR #115, opened 2015-10-13
Add manual time increment (as a more controlled alternative to `tick=True`)

aehlke:
It's useful to be able to use a frozen time and easily up-tick it manually, only as desired. Adds an `increment` method to the time freezer that takes an optional delta (defaulting to 1 second).

An example use case: in a common MySQL config, timestamps lack sub-second precision. With this, tests that do not mock time introduce a race condition where a quick series of timestamps may have the same value, or may straddle two seconds depending on when the test is ran. In this case, `tick=True` wouldn't help. This gives a nice shortcut to control time but still allow it to flow forward without having to entirely re-freeze.

### DIFF
--- a/freezegun/api.py
+++ b/freezegun/api.py
@@ -254,6 +254,9 @@ class FrozenDateTimeFactory(object):
     def __call__(self):
         return self.time_to_freeze
 
+    def tick(self, delta=datetime.timedelta(seconds=1)):
+        self.time_to_freeze += delta
+
 
 class _freeze_time(object):
 
@@ -323,13 +326,12 @@ class _freeze_time(object):
             return klass
 
     def __enter__(self):
-        self.start()
+        return self.start()
 
     def __exit__(self, *args):
         self.stop()
 
     def start(self):
-
         if self.tick:
             time_to_freeze = TickingDateTimeFactory(self.time_to_freeze, real_datetime.now())
         else:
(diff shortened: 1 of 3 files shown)


freezegun/api.py
```diff
@@ -429,6 +434,15 @@ def stop(self):
         time.localtime = time.localtime.previous_localtime_function
         time.strftime = time.strftime.previous_strftime_function
 
+    def increment(self, delta=datetime.timedelta(seconds=1)):
+        if self.tick:
+            raise TypeError("Cannot manually increment when tick is True.")
```
aehlke: Not happy with having this here, IMO the different frozen time classes should implement `increment` or not on their own, but that would require more extensive refactoring.

aehlke:
It seems the failing test is a false negative. It's complaining about syntax in a coverage.py file?

aehlke:
I updated this (and squashed commit) to have a proper API, so you can do:
```
    with freeze_time(initial_datetime) as frozen_datetime:
        frozen_datetime.tick()
```
(please see test for full example)

aehlke:
If someone can point me in a direction as to why the build is failing, I'll gladly work on a fix for it: https://travis-ci.org/spulec/freezegun/jobs/85430272 Not sure how to interpret this error.

spulec:
Thanks for this!

The test seems to be that newer Coverage versions don't support python 3.2. I've fixed in master.

spulec:
agreed

PR #121, opened 2015-12-10
Fix assertion in usage example

mre:
Also added missing module imports to make the example fully working.
Formatting is compatible with PEP8 now.


spulec:
Thanks!

PR #122, opened 2016-01-04
Clean up and speed up start().

agriffis:
This commit cleans up start() by removing repeated lists and the long
if-then chain. It's also faster (about 25%) by avoiding `if attribute_value
not in real_things:` which tests equality rather than object identity.

Avoiding the equality test has another bonus, which is that we can remove
the try/except. This works fine in my test environment, which includes
hiredis.


spulec:
thanks!

PR #125, opened 2016-01-08
Catch TypeError when fetching attribute values

ekampf:
On the latest cryptography 1.2 we started getting TypeError here.
Apparently one of the values when iterating dir(module) is a number and not a string :S


spulec:
thanks!

skakri:
@spulec when can we expect a new release? Not a big fan of outdated dependencies (i.e cryptography) nor istalling dependencies from git master. :)

allardhoeve:
Yes, a release would be appreciated.

PR #127, opened 2016-01-11
Wrap dir(module) in try-except block.

### LINKED ISSUE #126: freezegun crashes with cryptography 1.2 because of integer dict keys
Hello,

I originally filed this on cryptography, but they said it was probably more appropriate for freezegun or cffi.  Please let me know if you think it's more appropriate for cffi.  Thank you!

Cheers,
Fitz

Original bug report:
----

With the release of cryptography-1.2.0, my freezegun-using tests have started crashing.  The traceback looks like:

````
  File "breaker2.py", line 10, in <module>
    with freezegun.freeze_time('2015-10-31 12:00:01'):
  File "/Users/fitz/.virtualenvs/cgdb/lib/python3.4/site-packages/freezegun/api.py", line 321, in __enter__
    self.start()
  File "/Users/fitz/.virtualenvs/cgdb/lib/python3.4/site-packages/freezegun/api.py", line 374, in start
    for module_attribute in dir(module):
TypeError: unorderable types: str() < int()
````

The crash occurs on [this line](https://github.com/spulec/freezegun/blob/master/freezegun/api.py#L375) when freezegun attempts to `dir()` on `_cffi_backend._extern_py`.  That module has integer keys in its `__dict__` method, causing the above error.  The dump of the `__dict__` looks like this:

````
{4441887152: (<ctype 'void(*)(int, int, char *, int)'>,
              <bound method type._lock_cb of <class 'cryptography.hazmat.bindings.openssl.binding.Binding'>>,
              b'\x00\x00\x00\x00\x00\x00\x00\x00',
              None),
 4441887184: (<ctype 'int(*)(char *, int, int, void *)'>,
              <function _pem_password_cb at 0x1077e3488>,
              b'\x00\x00\x00\x00\x00\x00\x00\x00',
              None),
 4441887216: (<ctype 'int(*)(unsigned char *, int)'>,
              <function _osrandom_rand_bytes at 0x108b6f2f0>,
              b'\xff\xff\xff\xff\x00\x00\x00\x00',
              None),
 4441887248: (<ctype 'int(*)()'>,
              <function _osrandom_rand_status at 0x108b6f378>,
              b'\x00\x00\x00\x00\x00\x00\x00\x00',
              None),
 '__doc__': None,
 '__loader__': None,
 '__name__': '_cffi_backend._extern_py',
 '__package__': None,
 '__spec__': None}
````

The suspect commit is 830af271de1c9c358, when cryptography was updated to use cffi's static callbacks.  [Here is a sample script that demonstrates the breakage.](https://gist.github.com/felliott/4547366a649dac4df6d1)

I'm relatively new to Python, so I'm not sure if this is a bug in cffi, cryptography, or freezegun. I'm happy to refile elsewhere if needed.  Please let me know if you need any more information.  Thank you very much and have a great day!

Cheers,
Fitz

### PR DESCRIPTION
**Pull request by frgtn**
To prevent modules with incomparable keys from breaking freezegun
patching.

Fixes #126.

### DIFF
--- a/freezegun/api.py
+++ b/freezegun/api.py
@@ -372,7 +372,11 @@ class _freeze_time(object):
                 continue
             elif (not hasattr(module, "__name__") or module.__name__ in ('datetime', 'time')):
                 continue
-            for module_attribute in dir(module):
+            try:
+                module_attributes = dir(module)
+            except TypeError:
+                continue
+            for module_attribute in module_attributes:
                 if module_attribute in real_names:
                     continue
                 try:

### CODE REVIEW

**Comment by JanMalte** [comment]
Please merge this PR as the error won't be fixed in cffi (see #126 )

**Comment by eisensheng** [comment]
I would suggest to use the following snippet to fill ``module_attributes`` in order to fix this issue since it handles non string type keys more gracefully.
```python
[k for k in module.__dict__.keys() if isinstance(k, str)]
```

It's beyond me why cffi won't fix the issue. They're stretching the definition of module global keys.

spulec:
This appears to be fixed in `cffi > 1.5.2` according to #126

PR #131, opened 2016-01-31
do not use hex numbers in readme

aliva:
current readme suggest using using numbers like `01` for setting minutes, which:
* in python2 represents hex numbers (and 08 is invalid and throws `SyntaxError`)
*  python3 throws `SyntaxError` for 01

while a two digit number is really nice way to show minutes, to reduce misunderstanding please use a normal `int` type in readme :-)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -26,7 +26,7 @@ Decorator
 
       @freeze_time("2012-01-14")
       def test():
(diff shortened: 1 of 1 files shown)


spulec:
thanks!

Issue #134, opened 2016-02-14
Tarball on PyPI contains .pyc files


FedericoCeratto:
There are 25 .pyc files in freezegun-0.3.6.tar.gz

PR #139, opened 2016-03-03
Globally exclude __pycache__ and py[co] from sdist

koobs:
This change adds two global exclude patterns to `MANIFEST.in`, to preclude compiled/binary files from being added to the PyPI source distribution (sdist)

This change was tested to confirm the non-presence of these files in a newly generated sdist, by creating a new sdist using `python setup.py sdist`, using the sources from the existing tarball (that contain the bad files).

Resolves https://github.com/spulec/freezegun/issues/134


spulec:
thanks!

PR #145, opened 2016-06-25
Fixing halfdone work using None parameter for freeze_time()

xeor:
This is a small fix to support using `None` as the input-parameter for freeze_time(). This is very useful if you are moving time a lot, but don't care about the initial time.


spulec:
thanks!

Issue #99, opened 2015-06-29
freeze_time might leave some instances patched after exit


coagulant:
Module A gets imported inside `freeze_time` context manager
Module A imports `datetime.datetime` (patched with freezegun).
Context manager exists and unpatches all modules it has tracked in `undo_changes `.

The problem is that module A still has `FakeDatetime` instead of `datetime.datetime`,
which results in something like that if it gets called:
```
    return cls.times_to_freeze[-1]
IndexError: list index out of range
```

Sorry, I have no test to show this bug.

spulec:
Thanks for opening! I have recreated the issue.

This is a result of a performance improvement that was made with https://github.com/spulec/freezegun/commit/ac163a5c534b4728bd082d3f67d956b1df71f04c. Clearly we will need to rethink a better way to do that.

For now, you can try using an older version.

UberFarmer:
I was getting the same bug in some cases with Python 2.7, I tried downgrading to fix the issue. The problem was still happening on 0.2.3, but was fixed when I downgraded to 0.2.2. So I bet the issue is not due to ac163a5c534b4728bd082d3f67d956b1df71f04c, but rather a change between 0.2.2 and 0.2.3.

Once it's in the corrupted state, we can repro with calls to `datetime.datetime.utcnow()`, here's the partial stack trace from 0.2.3:
```
env/local/lib/python2.7/site-packages/freezegun/api.py:137: in utcnow
    result = cls._time_to_freeze()
env/local/lib/python2.7/site-packages/freezegun/api.py:142: in _time_to_freeze
    return cls.times_to_freeze[-1]
E   IndexError: list index out of range
```

spulec:
@UberFarmer  Do you have a test case for it breaking on 0.2.3? My test cases all pass on 0.2.5

(obviously we still need to fix in master too, but I just want to ensure I understand the issue fully)

UberFarmer:
Turns out my bug may be caused by a different set of events than @spulec , but the symptoms are the same once it happens.

I can't get a really nice repro for you, but here's some more info. When a certain set of dependencies are installed (not sure exactly which combination, but some are definitely related to werkzeug/flask), trying to test the following code causes freezegun 0.2.3+ to enter a corrupted state. Once it's in the corrupted state, it behaves as specified in my previous comment.

```lang=python
from freezegun import freeze_time
import unittest

class FreezegunTestCase(unittest.TestCase):

    @freeze_time('2015-01-01')
[...]

PR #155, opened 2016-09-28
Fix for #99 -- if no time to be frozen is available use the real datetime

mindojo-victor:


spulec:
Is this necessary? Should we ever be calling `_time_to_freeze` if there isn't `cls.times_to_freeze`?

mindojo-victor:
> Should we ever be calling `_time_to_freeze` if there isn't `cls.times_to_freeze`?

You mean to check if `cls.times_to_freeze` is empty in each place where we call `_time_to_freeze`?

spulec:
I'm trying to figure out the scenario in which this happens. Do you have a test to reproduce the issue?

mindojo-victor:
I don't have the reproducible tests. It happens sometimes. Looks like it depends on test sequence. Some tests patch the `datetime` module, then next `test_*` is module is run, but the decorator is already finished. I am using `nose2`.

If I were to decide what to do with this, I would not try to restore original `datetime` module. Just leave the module patched: if there is a frozen datetime to return, return it, otherwise return pass through to the original `datetime`. This would make the implementation much simpler.

spulec:
thanks!

AndreasBackx:
@spulec could a new release be published to Pypi that includes this and the other new changes? We're currently using the latest commit for testing because of this bug.

spulec:
Version `0.3.9` is now released!

PR #156, opened 2016-10-02
Register fake types in PyMySQL conversions

danqing:
PyMySQL creates the `conversion` object as a copy of the `encoders` object, so we need to register the type there as well. `Connection` for one uses `conversion`, so without this registration most SQL operations will fail.

Cf. https://github.com/PyMySQL/PyMySQL/blob/master/pymysql/converters.py#L417


spulec:
Thanks!

PR #160, opened 2016-10-24
Fix test to work when current timezone is GMT-14 or GMT+14.

### LINKED ISSUE #157: test_import_localtime fails in GMT-14
Hello and thanks for developing freezegun.
The following test is failing when the system is running it a GMT-14 timezone, maybe because the freeze time is being set to 12:00 

https://github.com/spulec/freezegun/blob/master/tests/test_class_import.py#L78

A failed build log is available at:
https://tests.reproducible-builds.org/debian/logs/unstable/amd64/freezegun_0.3.6-1.build2.log.gz

vincentbernat:
Wikipedia says that UTC+14 is a thing. /usr/share/zoneinfo says that
GMT-14 is a thing. Ensure both of them work in tests when using local
time. Python accepts up to a whole day (so, GMT+23 is valid). Fix the
test on local time to accept all that.

Fix #157.

### DIFF
--- a/tests/test_class_import.py
+++ b/tests/test_class_import.py
@@ -80,7 +80,8 @@ def test_import_localtime():
     struct = fake_localtime_function()
     assert struct.tm_year == 2012
     assert struct.tm_mon == 1
-    assert struct.tm_mday == 14
+    assert struct.tm_mday >= 13  # eg. GMT+14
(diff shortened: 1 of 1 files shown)


spulec:
Thanks!

PR #161, opened 2016-10-27
Fixed #101 - ignore warnings when patching imported modules.

### LINKED ISSUE #101: Freezetime's access to every imported module's attribute triggers celery's deprecation warnings
Since https://github.com/spulec/freezegun/commit/70518d630c71a37d3c651dee7b73d347b5d75405 freezegun has been fetching every attribute from every imported module to check if it matches one of the patchable date. However, the celery package imports a number of modules that throw deprecation warnings on their access.  Celery slightly complicates the tracking down of this issue because it implements some LazyModule functionality so the deprecation warnings show when first accessed. Normally this prevents the warnings from showing to developer unless they intended to use a deprecated module, but since freezegun touches ever attribute on every module, the warning is always thrown.

pelme:
freeze_time introspects imported modules in sys.modules and looks for
datetimes to replace. However, when calling `getattr()` or `hasattr()` on a module
attribute, warnings can be emitted. Regular Python modules cannot cause
this problem, this only happens in libraries that constructs modules
dynamically.

This commit captures and ignores warnings that happens during
freeze_time `start()` and `stop()` introspection while not affecting warnings emitted by regular
code using freeze_time.

Prominent libraries that utilize this technique includes py.code (used by pytest) and celery:

py.code - the compiler package import causes a warning to be emitted:
 lazy loading: https://github.com/pytest-dev/py/blob/67987e26aadddbbe7d1ec76c16ea9be346ae9811/py/__init__.py
 warning: https://github.com/pytest-dev/py/blob/67987e26aadddbbe7d1ec76c16ea9be346ae9811/py/_code/_assertionold.py#L3

celery.task - the sets module emits a warning:
 lazy loading: https://github.com/celery/celery/blob/46c92025cdec07a4a30ad44901cf66cb27346638/celery/task/__init__.py
 warning: https://github.com/celery/celery/blob/46c92025cdec07a4a30ad44901cf66cb27346638/celery/task/sets.py

I've manually verified that this patch fixes the celery and py.code issue in my own code base.

### DIFF
--- a/freezegun/api.py
+++ b/freezegun/api.py
@@ -6,6 +6,7 @@ import time
 import calendar
 import unittest
 import platform
+import warnings
 
 from dateutil import parser
 from dateutil.tz import tzlocal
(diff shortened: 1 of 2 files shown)


spulec:
Excellent, thank you!

PR #163, opened 2016-10-27
Fixed #162 - allow decorating old-style classes.

### LINKED ISSUE #162: Decorating old style classes does not work
https://github.com/spulec/freezegun/commit/4b4ba0e2a14c1ecd7e8bfc0b2ea1f332840f5568 changed the class decorator to use `.mro()` which broke the support for decorating an old style class in Python 2.

Just having plain classes and omitting object is a common pattern in pytest tests:

``` python
@freeze_time('2015-12-24')
class Test_Something:
    def test_foo(self):
        ...
```

This is not a problem in Python 3 since there are no old style classes. I have a patch which addresses this, will soon publish a PR.

pelme:
See #162 for more information. This commit fixes a regression which broke support for python 2 old-style classes.

### DIFF
--- a/freezegun/api.py
+++ b/freezegun/api.py
@@ -329,7 +329,9 @@ class _freeze_time(object):
         else:
 
             seen = set()
-            for base_klass in klass.mro():
+
+            klasses = klass.mro() if hasattr(klass, 'mro') else [klass] + list(klass.__bases__)
+            for base_klass in klasses:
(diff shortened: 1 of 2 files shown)


pelme:
This PR originally had a broken fix, it is now amended and should be correct!

spulec:
Great, thank you!

PR #169, opened 2016-12-21
Fix README styling to valid reStructuredText

aleksihakli:
Hey,

I noticed that the PyPI page looks strange:

https://pypi.python.org/pypi/freezegun

The README reStructuredText was broken so I made some fixes, and they seem to work on at least [ninjs](http://rst.ninjs.org/?n=194db2b974877802bc411365a5242883&theme=basic) editor, which highlighted the broken titles earlier.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -19,28 +19,28 @@ Decorator
 
 .. code-block:: python
 
-      from freezegun import freeze_time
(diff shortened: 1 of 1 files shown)


spulec:
Thanks!

PR #172, opened 2017-02-06
Add monkey-patching to the uuid module (uuid1)

romuald:
The uuid.uuid1() method is using the time() function, but is not
consistent on every platform since it **may** use the ctypes methods if
available.

"Fix" this by patching the uuid module internal references so it will
always use the python version, thus making uuid1() method always return
the uuid from the frozen date

### DIFF
--- a/freezegun/api.py
+++ b/freezegun/api.py
@@ -3,6 +3,7 @@ import functools
 import inspect
 import sys
 import time
+import uuid
 import calendar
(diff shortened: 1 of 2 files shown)


romuald:
I've created a new test module since I'm not sure where the correct place was

spulec:
Great, thank you!

Issue #69, opened 2014-12-29
0.2.3 performance issue


matino:
Hi,

I run into a performance issue with 0.2.3 version. 0.2.2 works just fine.
Below the sample from profiler output for test function. Also the profiler shows muchbigger output for 0.2.3.

``` Python
from freezegun import freeze_time
import datetime
def test():
    with freeze_time('2014-10-10'):
        print datetime.datetime.now()
```

**0.2.3**:

```
ncalls  tottime  percall  cumtime  percall filename:lineno(function)
1    0.010    0.010    0.395    0.395 api.py:264(stop)
1    0.009    0.009    0.448    0.448 api.py:223(start)
1    0.000    0.000    0.395    0.395 api.py:220(__exit__)
1    0.000    0.000    0.448    0.448 api.py:217(__enter__)
```

**0.2.2**:

```
ncalls  tottime  percall  cumtime  percall filename:lineno(function)
1    0.001    0.001    0.003    0.003 api.py:223(start)
1    0.001    0.001    0.002    0.002 api.py:255(stop)
1    0.000    0.000    0.002    0.002 api.py:220(__exit__)
1    0.000    0.000    0.003    0.003 api.py:217(__enter__)
```

simon-weber:
I ran into this too. For anyone who can't wait for a fix, I wrote a drop-in alternative that's much faster (2 orders of magnitude on my machine) but only supports a subset of what freezegun does: https://github.com/simon-weber/python-libfaketime.

Some ideas on making freezegun faster:
* make the new behavior optional:
  * easiest fix I can think of
  * doesn't fix the problem for people who need aliased module support, though having folks find-and-replace is probably easier than adding freezegun support :grin: 
* [abuse the garbage collector](https://docs.python.org/2/library/gc.html#gc.get_referrers) to make the search for aliased modules faster: 
  * potentially a quick fix
  * I'm not sure if a) it'd work, or b) it'd even be much faster (since python probably doesn't keep a reverse index for all objects?)
  * pretty ridiculous
* change the replacement logic to install patches once, then leave them in place and dynamically turn them on and off:
  * probably the most robust solution, but might take some time to implement
  * introduces some overhead on non-mocked date/time calls. https://pypi.python.org/pypi/wrapt could alleviate some of that if needed
  * this is what https://github.com/venmo/btnamespace does ([code](https://github.com/venmo/btnamespace/blob/master/btnamespace/namespace.py))

Hope that helps!

spulec:
[...]

PR #175, opened 2017-03-06
Implement cache meachanism to speed up module search

hltbra:
This should fix the performance issues reported in issue #69.

I'd like to hear from other people that experienced the slowdowns to see if this pull request will fix their issues.

cc @Deimos

### DIFF
--- a/freezegun/api.py
+++ b/freezegun/api.py
@@ -1,5 +1,6 @@
 import datetime
 import functools
+import hashlib
 import inspect
 import sys
 import time
@@ -18,6 +19,8 @@ real_gmtime = time.gmtime
 real_strftime = time.strftime
 real_date = datetime.date
 real_datetime = datetime.datetime
+real_date_objects = [real_time, real_localtime, real_gmtime, real_strftime, real_date, real_datetime]
+
 
 try:
     real_uuid_generate_time = uuid._uuid_generate_time
@@ -33,6 +36,60 @@ try:
     import copy_reg as copyreg
 except ImportError:
     import copyreg
+
+
+# keep a cache of module attributes otherwise freezegun will need to analyze too many modules all the time
+_GLOBAL_MODULES_CACHE = None
+
+
+def _get_global_modules_cache():
+    global _GLOBAL_MODULES_CACHE
+    if _GLOBAL_MODULES_CACHE is None:
+        _GLOBAL_MODULES_CACHE = _setup_modules_cache()
+    return _GLOBAL_MODULES_CACHE
+
+
+def _get_module_attributes(module):
+    result = []
+    for attribute_name in dir(module):
+        try:
+            attribute_value = getattr(module, attribute_name)
+        except (ImportError, AttributeError, TypeError):
+            # For certain libraries, this can result in ImportError(_winreg) or AttributeError (celery)
+            continue
+        else:
+            result.append((attribute_name, attribute_value))
+    return result
+
+
+def _setup_modules_cache():
+    # FIXME: move this definition to be at the top-level
+    real_time_object_ids = set(id(obj) for obj in real_date_objects)
+    result = {}
+    for mod_name, module in sys.modules.items():
+        # ignore this module
+        if mod_name == __name__:
+            continue
+        date_attrs = []
+        all_module_attributes = _get_module_attributes(module)
+        for attribute_name, attribute_value in all_module_attributes:
+            if id(attribute_value) in real_time_object_ids:
+                date_attrs.append((attribute_name, attribute_value))
+        result[mod_name] = (_get_module_attributes_hash(module), date_attrs)
+    return result
+
+
+def _get_module_attributes_hash(module):
+    return '{}-{}'.format(id(module), hashlib.md5(','.join(dir(module))).hexdigest())
+
+
+def _get_cached_module_attributes(mod_name, module):
+    global_modules_cache = _get_global_modules_cache()
+    module_hash, cached_attrs = global_modules_cache.get(mod_name, ('0', []))
+    if _get_module_attributes_hash(module) == module_hash:
+        return cached_attrs
+    else:
+        return _get_module_attributes(module)
 
 
 # Stolen from six
@@ -297,38 +354,6 @@ class FrozenDateTimeFactory(object):
         delta = target_datetime - self.time_to_freeze
         self.tick(delta=delta)
 
-import hashlib
-
-def setup_modules_cache():
-    to_patch = {
-        'real_date': id(real_date),
-        'real_datetime': id(real_datetime),
-        'real_gmtime': id(real_gmtime),
-        'real_localtime': id(real_localtime),
-        'real_strftime': id(real_strftime),
-        'real_time': id(real_time),
-    }
-    result = {}
-    for mod_name, module in list(sys.modules.items()):
-        attrs = []
-        for module_attribute in dir(module):
-            if mod_name == __name__:
-                continue
-            try:
-                attribute_value = getattr(module, module_attribute)
-            except (ImportError, AttributeError, TypeError):
-                # For certain libraries, this can result in ImportError(_winreg) or AttributeError (celery)
-                continue
-            if id(attribute_value) in to_patch.values():
-                attrs.append(attribute_value)
-
-        result[mod_name] = (id(module), hashlib.md5(','.join(dir(module))).hexdigest(), attrs)
-    return result
-
-
-GLOBAL_MODULES_CACHE = None
-
-
 
 class _freeze_time(object):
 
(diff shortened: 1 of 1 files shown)


Deimos:
Works without any modifications necessary now, results seem about the same as I mentioned in there with the previous version - about 0.25s duration on the first test, and then it's difficult to tell if there's any change to other ones. Seems to just add about 0.25s to the run time in total, from a few unscientific runs.

freezegun/api.py
```diff
@@ -35,6 +38,60 @@
...
+    return '{0}-{1}'.format(id(module), hashlib.md5(','.join(dir(module)).encode('utf-8')).hexdigest())
+
+
+def _get_cached_module_attributes(mod_name, module):
+    global_modules_cache = _get_global_modules_cache()
+    module_hash, cached_attrs = global_modules_cache.get(mod_name, ('0', []))
+    if _get_module_attributes_hash(module) == module_hash:
+        return cached_attrs
+    else:
+        return _get_module_attributes(module)
```
spulec: Do we want to be setting these back on `global_modules_cache` to update the cache?

freezegun/api.py
```diff
@@ -35,6 +38,60 @@
...
+    return result
+
+
+def _setup_modules_cache():
+    # FIXME: move this definition to be at the top-level
+    real_time_object_ids = set(id(obj) for obj in real_date_objects)
+    result = {}
+    for mod_name, module in list(sys.modules.items()):
+        # ignore this module
+        if mod_name == __name__:
```
hltbra: @spulec: in the `_freeze_time.start` method I saw a similar logic (`if mod_name is None or module is None`) but I didn't understand the `None` checks and didn't put them here. When are they `None`?

freezegun/api.py
hltbra: Sure.

freezegun/api.py
hltbra: answering myself: module is None when it's a freezegun module.

hltbra:
I will test this against `moto` later. I think there's a bug after running this against a work (private) project.

nolanlum:
Any update on merging this PR? I'm also interested in the speedups this provides, but I'm running into a strange bug with the module cache:

```
module = <AliasModule 'py.error' for 'py._error.error'>

    def _setup_module_cache(module):
        global _GLOBAL_MODULES_CACHE
        date_attrs = []
        all_module_attributes = _get_module_attributes(module)
        for attribute_name, attribute_value in all_module_attributes:
            if id(attribute_value) in _real_time_object_ids:
                date_attrs.append((attribute_name, attribute_value))
>       _GLOBAL_MODULES_CACHE[module.__name__] = (_get_module_attributes_hash(module), date_attrs)
E       AttributeError: __name__

/usr/local/lib/python2.7/site-packages/freezegun/api.py:85: AttributeError

File: freezegun/api.py [hunk 3]
```diff
@@ -35,6 +39,66 @@
...
+
+
+def _setup_module_cache(module):
+    global _GLOBAL_MODULES_CACHE
+    date_attrs = []
+    all_module_attributes = _get_module_attributes(module)
+    for attribute_name, attribute_value in all_module_attributes:
+        if id(attribute_value) in _real_time_object_ids:
+            date_attrs.append((attribute_name, attribute_value))
+    _GLOBAL_MODULES_CACHE[module.__name__] = (_get_module_attributes_hash(module), date_attrs)
```
azmeuk: I encountered this bug:
```
local.virtualenv/lib/python2.7/site-packages/freezegun/api.py:557: in wrapper
    with self:
local.virtualenv/lib/python2.7/site-packages/freezegun/api.py:434: in __enter__
    return self.start()
local.virtualenv/lib/python2.7/site-packages/freezegun/api.py:490: in start
    module_attrs = _get_cached_module_attributes(mod_name, module)
local.virtualenv/lib/python2.7/site-packages/freezegun/api.py:96: in _get_cached_module_attributes
    global_modules_cache = _get_global_modules_cache()
local.virtualenv/lib/python2.7/site-packages/freezegun/api.py:53: in _get_global_modules_cache
    _setup_modules_cache()
local.virtualenv/lib/python2.7/site-packages/freezegun/api.py:75: in _setup_modules_cache
    _setup_module_cache(module)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

module = <AliasModule 'py.error' for 'py._error.error'>

    def _setup_module_cache(module):
        global _GLOBAL_MODULES_CACHE
        #if not hasattr(module, "__name__"):
        #    return
    
        date_attrs = []
        all_module_attributes = _get_module_attributes(module)
        for attribute_name, attribute_value in all_module_attributes:
            if id(attribute_value) in _real_time_object_ids:
                date_attrs.append((attribute_name, attribute_value))
>       _GLOBAL_MODULES_CACHE[module.__name__] = (_get_module_attributes_hash(module), date_attrs)
E       AttributeError: __name__

all_module_attributes = []
date_attrs = []
module     = <AliasModule 'py.error' for 'py._error.error'>

local.virtualenv/lib/python2.7/site-packages/freezegun/api.py:88: AttributeError
```
It seems ok if I add the following code at the beginning of the function:
```python
def _setup_module_cache(module):
    if not hasattr(module, "__name__"):
        return
    ...
```

azmeuk:
@hltbra @spulec I am very interested in improving the performances of freezegun. Is there something I can do to help with this PR?

spulec:
Thanks!

freezegun/api.py
```diff
@@ -37,6 +41,70 @@
...
+
+
+def _get_module_attributes_hash(module):
+    return '{0}-{1}'.format(id(module), hashlib.md5(','.join(dir(module)).encode('utf-8')).hexdigest())
```
Chronial: In line 63, you catch a `TypeError` when you call `dir(module)`, here you don't. If a `TypeError` is raised there, that doesn't seem to prevent this function from being called. Is that an issue?
Why don't you pass the already existing `all_module_attributes` to this function and extract the names from that?

spulec: Thanks, fixed with https://github.com/spulec/freezegun/commit/689a7bbca2f1d39e5e35c83664f5e6fc73b07f40.

freezegun/api.py
```diff
@@ -66,7 +66,7 @@ def _setup_modules_cache():
     # FIXME: move this definition to be at the top-level
     real_time_object_ids = set(id(obj) for obj in real_date_objects)
     result = {}
-    for mod_name, module in sys.modules.items():
+    for mod_name, module in list(sys.modules.items()):
```
mariuszthecoder: it's unnecessary - if `sys.modules` is dict , then `sys.modules.items()` is iterable containing pairs of key and value in both py2 and py3  
casting it to `list` will not break it, but slow down a bit

spulec: Without creating the copy, I'm occasionally getting a `RuntimeError: dictionary changed size during iteration`, which makes sense since additional modules can get loaded.

Ideas?

Chronial: I would assume that the size of that collection is small enough so that casting to list once does not cause a relevant slowdown.

mariuszthecoder: so commit message is misleading

Chronial: No, the PR title is about an architectural optimization that saves seconds of run time, not microoptimizations that make a difference of a few microseconds.

pzelnip:
With 0.3.10 over 0.3.9 I'm seeing about a 16% speedup.  Not bad.

freezegun/api.py
mariuszthecoder: I meant this particular commit, not entire PullRequest  
Reading `Make sys.modules.items() compatible with python 3` I expect some py3 adjustment, but I get runtime error prevention.  
That was my point, sir :)

mariuszthecoder:
How do you measure speed improvement?

pzelnip:
I ran the tests and looked at the time that nose said it took to run them.  Did that 4 times and took the average.  Then uninstalled 0.3.9 & installed 0.3.10 and did the same.  The difference between the two averages was about 16%.

freezegun/api.py
Chronial: The commit message is not misleading. In python 2, `items()` returns a list (copy), which is what is needed here. But in python 3 it returns an iterator, which causes faulty behavior. So it is casted to list for python 3 support.

mariuszthecoder: +1

PR #184, opened 2017-05-29
Fix ignored modules

remcohaszing:
A check was made for each module if its name starts with the ignore
tuple. No string ever starts with a tuple. Now it is checked for every
value inside of the ignore tuple if the module name starts with this
value.

### DIFF
--- a/freezegun/api.py
+++ b/freezegun/api.py
@@ -414,7 +414,7 @@ class _freeze_time(object):
             for mod_name, module in list(sys.modules.items()):
                 if mod_name is None or module is None:
                     continue
-                elif mod_name.startswith(self.ignore):
+                elif any(map(mod_name.startswith, self.ignore)):
                     continue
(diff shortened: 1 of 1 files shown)


korkmaz:
@remcohaszing `startswith` already does that.

see the documentation.
```
str.startswith(prefix[, start[, end]])
Return True if string starts with the prefix, otherwise return False. prefix can also be a tuple of prefixes to look for. With optional start, test string beginning at that position. With optional end, stop comparing string at that position.

Changed in version 2.5: Accept tuples as prefix.
```
https://docs.python.org/2/library/stdtypes.html#str.startswith
https://docs.python.org/3/library/stdtypes.html#str.startswith

spulec:
Yes, I think this change isn't necessary. Are you seeing any issues without the change?

remcohaszing:
I must have done something wrong when I encountered this issue. Can't reproduce it myself.

PR #185, opened 2017-05-29
Add google.gax to default ignore list

remcohaszing:
google.gax depends on time.time() for communicating with external
systems. This is supposed to fix issues similar to
https://github.com/GoogleCloudPlatform/google-cloud-python/issues/3284


spulec:
I'm debating if we should find a more extensible way to do this. The idea of maintaining a list of all modules that might need real time modules doesn't seem ideal.

Issue #182, opened 2017-05-22
Should freeze_time accept lambdas or generators?


azmeuk:
Hi. I would love to see the following code work:
```python
import freezegun, datetime

dates = (datetime.datetime(year, 1, 1) for year in range(1, 2000))
with freezegun.freeze_time(lambda: next(dates)):
    print(datetime.datetime.now()) # would print the first of January "1
    print(datetime.datetime.now()) # would print the first of January "2
```
The idea is that in some test one may want that each `datetime.now` call return a different value he decided himself. This code is a simple example, but we can easily imagine lambdas more complicated.

What do you think? Would you accept pull requests for such a feature?

rodude123:
I think it should accept both I have tried both and both work fine

azmeuk:
@rodude123 The sample of code I provided raises this exception: `TypeError: freeze_time() expected None, a string, date instance, or datetime instance, but got type <type 'function'>.`

spulec:
If it doesn't work, I'd be happy to accept a PR to make it work. It seems like a reasonable use case.

spulec:
Fixed with #187

PR #187, opened 2017-06-06
Handling functions and generator objects

azmeuk:
As mentioned in #182, this PR allow to use functions and generators with `freeze_time()`. So now the next pieces of code are valid:
```python
def test_lambda():
    with freeze_time(lambda: datetime.datetime(2012, 1, 14)):
        assert datetime.datetime.now() == datetime.datetime(2012, 1, 14)

def test_generator():
    datetimes = (datetime.datetime(year, 1, 1) for year in range(2010, 2012))

    with freeze_time(datetimes):
        assert datetime.datetime.now() == datetime.datetime(2010, 1, 1)

    with freeze_time(datetimes):
        assert datetime.datetime.now() == datetime.datetime(2011, 1, 1)

    # The next call to freeze_time(datetimes) would raise a StopIteration exception.
```
The generator part of this PR is mostly sugar, as we could manage to have the same behavior using lambdas, for example with `freeze_time(lambda: next(my_generator))` instead of `freeze_time(my_generator)`.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -93,10 +93,10 @@ FreezeGun uses dateutil behind the scenes so you can have nice-looking datetimes
     def test_nice_datetime():
         assert datetime.datetime.now() == datetime.datetime(2012, 1, 14)
 
-Function objects
-~~~~~~~~~~~~~~~~
+Function and generator objects
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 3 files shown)


spulec:
Awesome work, thank you!

dhui:
@spulec Any plans on pushing a new release to PyPi? I'd love to start using this functionality in our tests!

PR #192, opened 2017-06-16
Split datetimes generator to a new function: freeze_times

sblondon:
Perhaps it's better to limit the accepted parameter to freeze_time() to one element parameter (a string, a datetime, etc.) and when several elements are provided, use another function. I called it freeze_times(). 

The patch move the code of the generator to this new function.

However, it adds a second function into the API so you could consider it's more complicated for the user.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -93,10 +93,10 @@ FreezeGun uses dateutil behind the scenes so you can have nice-looking datetimes
     def test_nice_datetime():
         assert datetime.datetime.now() == datetime.datetime(2012, 1, 14)
 
-Function and generator objects
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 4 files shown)


spulec:
Thanks for opening this.

Can you talk more about the use case? It isn't clear to me why it makes more sense to have two entry points.

sblondon:
As the function generates several differents datetimes, I think it's more explicit with a plural name function.

There is no specific needs more than that.
Don't hesitate to refuse this patch if you're not convinced.

dhui:
If you'd like another data point, I prefer using the same name. e.g. allow `freeze_time()` to support generators.

spulec:
Thanks for opening, but I'm going to decline this because I think the additional complication isn't worth it.

PR #198, opened 2017-07-12
Allow timedeltas to be used for tz_offset

Syeberman:
This makes it easier to call `freeze_time` in a way that respects the existing `tzinfo` on the `datetime` object:

```python
freezegun.freeze_time(dt, tz_offset=dt.utcoffset())
```

It also improves handling of non-integer offsets like [Newfoundland Time Zone](https://en.wikipedia.org/wiki/Newfoundland_Time_Zone).  Using `tz_offset=3.5` currently works, but I'm always wary of rounding errors when it comes to floats.  With this change you can write:

```python
freeze_time("2012-01-14", tz_offset=-datetime.timedelta(hours=3, minutes=30))
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -82,6 +82,10 @@ Timezones
         # datetime.date.today() uses local time
         assert datetime.date.today() == datetime.date(2012, 1, 13)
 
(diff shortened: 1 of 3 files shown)


spulec:
Thanks!

Issue #142, opened 2016-04-20
Scale ticks to simulate hours/days passing automatically


thedrow:
I got that idea from [timecop](https://github.com/travisjeffery/timecop#timecopscale).
Basically it means that every tick will "travel" a predefined amount of time instead of just one second.
It's currently possible to do that manually but it'd be nice to be able to do so automatically as well.

spulec:
Agreed, would be happy to accept a PR.

sheepsy90:
Added pull request

spulec:
Closed with #201

PR #201, opened 2017-08-03
#142 Added AutoStepTimeFactory

sheepsy90:
Saw the feature request - needed it myself - implemented it - only drawback is the conflict with the tick parameter as there are now 3 different factories. Maybe a refactoring necessary to support multiple of those factories on init. That would also support generators being passed to have custom steps in time.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -128,6 +128,22 @@ parameters which will keep time stopped.
     def test_nice_datetime():
         assert datetime.datetime.now() > datetime.datetime(2020, 1, 14)
 
+``auto_tick_seconds`` argument
+~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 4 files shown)


sheepsy90:
@spulec What about this pull request - i kind of need the functionality i added a lot - might be useful for others as well

spulec:
Thanks for this.

If you resolve conflicts and undo the version change, I'll be happy to merge

languitar:
Any chance to get this rolling again?

boxed:
The CI build seems to fail because python 3.3 isn't available on travis? Seems like a safe thing to ignore that and merge this?

spulec:
Thanks!

PR #203, opened 2017-08-23
add optional support for patching tzlocal.get_localzone()

danchr:
We override it to always use the 'current' timezone, as a pytz object.
Support depends on whether tzlocal is installed.


spulec:
Looks good. If you fix the conflicts, I'll be happy to merge

PR #210, opened 2017-09-22
Added option to pass in the factory to a decorated class

terrycain:
I have a few examples in the Moto project where we have the following code which will freeze time for the mocked moto service its wrapping.
```python
@freeze_time("2016-01-01")
@mock_someservice
def test_something():
    # Do some stuff
```
We could do with being able to advance time whilst in the test case. So the patch provides the following functionality whilst still remaining backwards compatible
```python
@freeze_time("2016-01-01", as_arg=True)
@mock_someservice
def test_something(frozen_time):
    # Do some stuff
    frozen_time.move_to("2016-02-01")
    # Do more stuff
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -171,6 +171,13 @@ Freezegun allows moving time to specific dates.
             frozen_datetime.move_to(initial_datetime)
             assert frozen_datetime() == initial_datetime
 
+
+    @freeze_time("2012-01-14", as_arg=True)
+    def test(frozen_time):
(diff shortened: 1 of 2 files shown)


spulec:
Awesome, thanks!

tricard-eb:
When can we expect a release with that ?

spulec:
Version `0.3.10` has now been released.

tricard-eb:
awesome! thanks

PR #216, opened 2017-11-30
Fix uuid patch when jumping back in time

romuald:
The uuid module keeps an internal variable as some kind of cache that
prevents jumping back in time: the most recent timestamp would always be
returned


spulec:
Fixed with https://github.com/spulec/freezegun/pull/197

PR #219, opened 2017-12-21
Do not scan vendored copies of six.moves: Makes start faster

evanj:
Skip modules that end with the name .six.moves since many libraries
include their own copy. This makes starting freezegun much faster.
Freezegun already skips the module named "six.moves" by appending
it to the freezer's "ignore" list. However, many libraries ship their
own copy, and import it under their own name. For example, in
pkg_resources:

    from pkg_resources.extern import six

A list of copies of six.py in a project I work on shows the following
libraries include a copy, so I suspect this is pretty common:

boto
pip
pkg_resources
urllib3

### DIFF
--- a/freezegun/api.py
+++ b/freezegun/api.py
@@ -420,7 +420,7 @@ class _freeze_time(object):
             for mod_name, module in list(sys.modules.items()):
                 if mod_name is None or module is None:
                     continue
-                elif mod_name.startswith(self.ignore):
+                elif mod_name.startswith(self.ignore) or mod_name.endswith('.six.moves'):
                     continue
                 elif (not hasattr(module, "__name__") or module.__name__ in ('datetime', 'time')):
                     continue
--- a/freezegun/api.py
+++ b/freezegun/api.py
@@ -535,7 +535,6 @@ def freeze_time(time_to_freeze=None, tz_offset=0, ignore=None, tick=False, as_ar
     if ignore is None:
         ignore = []
     ignore.append('six.moves')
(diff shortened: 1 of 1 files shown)


evanj:
I used the following script to test this. To reproduce the problem, do:

1. `virtualenv bugvenv``
2. `bugvenv/bin/pip install freezegun`
3. `bugvenv/bin/python timings.py`

On my machine, without this change, I get the following timings:

```
BEFORE pkg_resources:
1000 iterations in 2.500854 s; 2.500854 ms / call
1000 iterations in 2.475269 s; 2.475269 ms / call
1000 iterations in 2.484204 s; 2.484204 ms / call

AFTER pkg_resources:
1000 iterations in 6.951676 s; 6.951676 ms / call
1000 iterations in 6.825846 s; 6.825846 ms / call
1000 iterations in 6.937931 s; 6.937931 ms / call
```

After this change, I get the following:

```
BEFORE pkg_resources:
1000 iterations in 2.555821 s; 2.555821 ms / call
1000 iterations in 2.595121 s; 2.595121 ms / call
1000 iterations in 2.513985 s; 2.513985 ms / call

AFTER pkg_resources:
1000 iterations in 4.956954 s; 4.956954 ms / call
1000 iterations in 5.005256 s; 5.005256 ms / call
1000 iterations in 5.007610 s; 5.007610 ms / call
```

In my real application's test suite, which has multiple copies of six, the difference is significantly more dramatic.

spulec:
Thanks!

PR #221, opened 2018-01-04
Asyncio/Coroutine support

### LINKED ISSUE #180: Mark notation does not work when using pytest-asyncio library
Simple test case fails when using both libraries and the mark notation:

```
import datetime
import asyncio
import pytest
import pytest_asyncio

@freeze_time("2012-01-14")
async def test():
    assert datetime.datetime.now() == datetime.datetime(2012, 1, 14)  # passes

@pytest.mark.asyncio
@freeze_time("2012-01-14")
async def test_async():
    assert datetime.datetime.now() == datetime.datetime(2012, 1, 14)  # fails
```

If we use a different notation then it works as expected:

```
import datetime
import asyncio
import pytest
import pytest_asyncio

@freeze_time("2012-01-14")
async def test():
    assert datetime.datetime.now() == datetime.datetime(2012, 1, 14)  # passes

@pytest.mark.asyncio
async def test_async():
    with freeze_time("2012-01-14"):
        assert datetime.datetime.now() == datetime.datetime(2012, 1, 14) # passes
```

Dependencies using Python 3:
```
freezegun==0.3.8
pytest==3.0.7
pytest-asyncio==0.5.0
```

ojii:
This patch fixes #180 but is not limited to pytest-asyncio. It adds support for using `freeze_time` as a decorator on coroutines (`@asyncio.coroutine` or `async def`).

### DIFF
--- a/freezegun/_async.py
+++ b/freezegun/_async.py
@@ -1,0 +1,17 @@
+import functools
+
+import asyncio
+
+
+def wrap_coroutine(api, coroutine):
+    @functools.wraps(coroutine)
+    @asyncio.coroutine
+    def wrapper(*args, **kwargs):
+        with api as time_factory:
+            if api.as_arg:
+                result = yield from coroutine(time_factory, *args, **kwargs)
+            else:
+                result = yield from coroutine(*args, **kwargs)
+        return result
+
+    return wrapper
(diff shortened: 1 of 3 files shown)


spulec:
This is great; thanks!

PR #223, opened 2018-01-24
Fix Module Iteration / Race Condition

lskillen:
First of all, I just wanted to say that we love freezegun - It's simplified a lot of our (rather complicated) testing around assertions for datetime-based functionality.

We did however encounter issues when combined with certain modules, like [py](https://pypi.python.org/pypi/py/1.5.2), which is dragged in by [tox](https://pypi.python.org/pypi/tox/2.9.1) -  The module iteration is broken within freezegun because `py.test` isn't a quite a module, so a `TypeError` is raised when calling `dir(module)` on it.

We also encountered a race condition on `tz_offsets` not being setup correctly in time for usage elsewhere, so this moves the logic towards the start. This occurs while navigating large trees of modules and attributes. This absolutely doesn't fix all race conditions, but it greatly minimises this specific scenario.

No regressions occurred in the test suite.

Thank you!

### DIFF
--- a/freezegun/api.py
+++ b/freezegun/api.py
@@ -381,7 +381,13 @@ class _freeze_time(object):
 
         # Change the modules
         datetime.datetime = FakeDatetime
+        datetime.datetime.times_to_freeze.append(time_to_freeze)
+        datetime.datetime.tz_offsets.append(self.tz_offset)
+
         datetime.date = FakeDate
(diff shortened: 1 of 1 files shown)


lskillen:
I suppose that means we'll need to invent a test for the TypeError. :-)

spulec:
Thanks!

PR #224, opened 2018-01-24
Optimize api.

mariuszthecoder:
Making optimizations I've multiplied test_simple_api 1000 times and measured execution times.
Execution time dropped from ~50s to ~38s (on localhost, py3.6) so it's 20 up to 25% faster.


spulec:
I'm going to close this in favor of https://github.com/spulec/freezegun/pull/175.

If there are still additional parts of this we should add, feel free to rebase and I can take another look.

mariuszthecoder:
ok, I will check it again then

PR #231, opened 2018-03-06
3.7 support (uuid module changes)

ojii:
Python 3.7 removed `uuid._uuid_generate_time`. It now has `uuid._load_system_functions` and `uuid._generate_time_safe`. `_generate_time_safe` is set by calling `_load_system_functions` (subsequent calls to that function are no-op). This change detects the missing `uuid._uuid_generate_time` attribute and uses the new attribute/function if they're missing.


spulec:
Awesome, thanks!

PR #232, opened 2018-03-07
_GLOBAL_MODULES_CACHE: Update on demand only

evanj:
Remove _get_global_modules_cache and _setup_modules_cache which
attempted to pre-load the modules cache for all modules. This is
unnecessary since _get_cached_module_attributes is the only function
that uses the cache, and it will update the cache if the values are
out of date or missing.

_get_cached_module_attributes: Use module.__name__ instead of passing
in mod_name. This is the attribute used by _setup_module_cache.

This also caused a problem with vendored copies of six.moves. Commit
7af9f039ab merged in 590504c9b9 added code to ignore any copy of
six.moves which was not respected by _setup_modules_cache.
Additionally, it did not ignore the self.ignore list. This change
avoids needing to keep these two module ignore lists in sync.

Without this change, version 0.3.10 takes an incredible amount of
time to run on Bluecore's code base, due to the same problem as the
original six.moves fix: Many libraries vendor six.moves, which
references other modules, which takes forever to scan. See:
https://github.com/spulec/freezegun/pull/219

### DIFF
--- a/freezegun/api.py
+++ b/freezegun/api.py
@@ -57,18 +57,7 @@ except AttributeError:
 
 
 # keep a cache of module attributes otherwise freezegun will need to analyze too many modules all the time
-# start with `None` as the sentinel value.
-# if `{}` (empty dict) was the sentinel value, there's a chance that `setup_modules_cache()` will be called many times
-_GLOBAL_MODULES_CACHE = None
-
-
-def _get_global_modules_cache():
-    global _GLOBAL_MODULES_CACHE
-    # the first call to this function sets up the global module cache. it's expected to be slower than consecutive calls
-    if _GLOBAL_MODULES_CACHE is None:
-        _GLOBAL_MODULES_CACHE = {}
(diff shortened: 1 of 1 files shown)


evanj:
This is not the only possible fix. We could also move the "module ignore logic" into a shared function that could be called by both `_setup_modules_cache` and `_freeze_time.start`. There may also be other fixes.

Without this, our test suite takes about ~45 minutes. With this, it takes about ~12 minutes. Probably because we use `nosetests` with the `--process-restartworker` option, so each test runs in its own process, so the cache is pretty much unused for us

spulec:
Thanks!

amw:
This is a great fix. Too bad it's not included in the most recent release – I've spent a couple hours today debugging forking errors on macOS (due to `six.moves` not being properly ignored) and only after I came up with the same fix I have found out that it's already been in `master` for a long time.

@spulec any idea when 0.3.11 will be released?

Issue #228, opened 2018-02-15
freeze_time has no effect on time.clock()


gathogojr:
When you apply `@freeze_time(datetime.datetime.now())` attribute to a unit test, time is frozen if a method being invoked by the unit test uses `time.time()` but not if `time.clock()` is used

**Illustration**:
```python
import time
import datetime
from django.test import TestCase
from freezegun import freeze_time


def method1():
    start = time.clock()
    # a set of operations here
    time.sleep(0.1)
    end = time.clock()
    duration = end - start  # Doesn't always yield zero when method1 is invoked from unit test
    print duration


def method2():
    start = time.time()
    # a set of operations here
    time.sleep(0.1)
    end = time.time()
    duration = end - start  # Yields zero when method2 is invoked from unit test
    print duration


class Temp(TestCase):
    @freeze_time(datetime.datetime.now())
    def test_temp(self):
        method1()
        method2()
```

Am I missing something? I'd be eager to know if anybody has experienced similar behaviour

colinfike:
I'm not sure I even know what the desired functionality is in this case. Since `time.clock` returns a float representing the processor time for your process since it began running (at least on Unix) would we just freeze that to `0.0`?

spulec:
Closed with https://github.com/spulec/freezegun/pull/239

PR #239, opened 2018-04-21
Added support for clock() function of time.

rinslow:
So basically this was done as requested in https://github.com/spulec/freezegun/issues/228

The author did not specify the desired functionality so I had to suggest my own, 
It's pretty clear from the unit-test

```python
def test_time_clock():
    with freeze_time('2012-01-14 03:21:34'):
        assert time.clock() == 0

        with freeze_time('2012-01-14 03:21:35'):
            assert time.clock() == 1

        with freeze_time('2012-01-14 03:21:36'):
            assert time.clock() == 2
```

once inside a freeze you can reset the CPU time given to the process with additional freezes.
Looking forward for your feedback :)


spulec:
Thanks; this looks great!

PR #249, opened 2018-06-05
Fix typo in long_description

hugovk:
There's no description on PyPI:

image

https://pypi.org/project/freezegun/#description

And only because of a little typo in setup.py!


spulec:
Thank you!

PR #260, opened 2018-09-10
Fixed nesting bug for time.time() and massive performance improvements

boxed:


boxed:
Yes, coverage will decrease if you optimize away redundant code :P

boxed:
@spulec ping? It would be nice to get this merged, we have had to introduce some hacky workarounds in our tests for this.

boxed:
test_ticking_time_clock does not pass with the latest commit. I have stared at it a long time and I can't understand how it could be correct. To me the line `assert second > first + 1` implies that the nested freeze to 2012-01-14 03:21:35 should inherit some of the time offset already accumulated from the previous outer frozen time. This seems wrong. I think that assert (and `assert third > second + 1`) worked precisely because nesting freezes was slow. So since I made nested freezes faster than the initial monkey patching freezing this test failed. I think those two asserts should just be removed. What do you think?

boxed:
To test the nesting time performance I have used the following script:

```python
import freezegun

freezer = freezegun.freeze_time(tick=True)
freezer.start()

for _ in range(1000):
    with freezegun.freeze_time():
        pass

freezer.stop()
```

before the last commit this script took 5.6-6.1 seconds on my machine. After that commit it takes ~0.61 seconds. This should be compared to ~0.56 seconds for the following script:

```python
import freezegun

for _ in range(1000):
    pass
```

@spulec

benkuhn:
Is it possible that you could fix the backwards incompatibility in the last commit by

1. Using `inspect.currentframe()` and traversing `frame.f_back` instead of using `extract_stack` to get the raw frame objects;
2. Determining the module name by `frame.f_globals["__name__"]`?

boxed:
Not really:

```
In [3]: %timeit traceback.extract_stack()
198 µs ± 2.41 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [4]: %timeit inspect.currentframe()
443 ns ± 2.92 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

So without even traversing a single frame I'm already doubling the time of the part of the code that contains the vast majority of the overhead, even after my change. The test suite I'm testing on takes ~25s and extract_stack is already ~8s. 

Before the changes in this PR it was 8 _minutes_ :P

boxed:
You are absolutely right, and that change makes it even faster. Super great suggestion! Now my test scenario is down to below 20 seconds and freezegun hardly even registers in the profiling output.

spulec:
This looks great; thanks for all the work!

Issue #269, opened 2018-10-17
The new way of ignoring module since 0.3.11 has broken pytest-freezegun


bdauvergne:
pytest-freezegun is adding the module _pytest module to ignorable module, since 0.3.11 module are loked up by following the stack trace on five level, it happend that coming from a pytest test you reach the _pytest module, and so any call to time.time() for example is redirected to the real function and not the freezed one.

bdauvergne:
There should be a way to indicate that a module must be ignored when he calls time(), but not in function called from it and not in the same module.

boxed:
This is my bad. I introduced it and haven't seen any ill effects on the test suite at work which uses pytest. Maybe the ignore rule should just be more narrowly defined to hit the specific places in pytest where it's relevant?

bdauvergne:
The bug is in, new API will not be integrated in pytest-freezegun immediately, in the mean time I think that you need to handle _pytest specially as you said.

bdauvergne:
Reading the code of _should_use_real_time() I find the way of matching module names too lax `module_name.startswith(ignore_list[-1])` I think it should be something like `module_name in ignore_lists[-1] or module_name.startswith([name + '.' for name in ignore_list[-1])`.

boxed:
I misremembered... I thought I had added _pytest to the ignore list, but that wasn't the case. You say something has changed, but the ignore behavior of using startswith is the same as it has been before this release so that doesn't seem to be the problem.

bdauvergne:
I'm not the developper from pytest-freezegun, what changed is that now you traverse the stack for 5 levels before deciding if a time() call should use the real function or the fake one, before you just changed the global reference from module target by the ignore list, it's a total change of behaviour. As for the startswith it's just that it's totally non precise (startswith.('whatever') will match whatever, whavever.* but also whateverxyz*, and I do not think that it is the wanted behaviour, the wanted behaviour is the package whatever or any package under it.

bdauvergne:
Is there some pilot in the plane ? :)

boxed:
@bdauvergne what does that mean?

bdauvergne:
#245 was a pull request provided by @kylegibson and accepted by @spulec, I don't know who I'm talking to and who could accept by issue as a real problem (as I'm sure there is one, as it is freezegun does not work anymore with pytest, which is clearly a big target as the most used test runner for now).

boxed:
I have commit access so can fix bugs. 

[...]

PR #270, opened 2018-10-17
add special handling of _pytest to prevent bug in pytest-freezegun #269

bdauvergne:
It passes current (under py27 and py36, I do not have other versions), and it makes my pytest-freezegun using tests working.


boxed:
I'll do https://github.com/spulec/freezegun/pull/274 instead.

PR #271, opened 2018-10-18
default to ignoring selenium

RaphaelKimmig:


boxed:
This seems like a very broad pattern... in https://github.com/spulec/freezegun/pull/270 we discuss a similar thing with pytest and just adding pytest as a whole sometimes included the test code in the ignore list which is bad. 

Are you sure selenium isn't ever a part of the stack trace when the code under test is run?

RaphaelKimmig:
I'm not 100% sure, but I would think that it isn't. Given that selenium is not a test runner, but rather controls a remote browser instance that tests against a running server. I have to admit that I haven't used selenium with python web frameworks other than django, but I assume the execution model is going to be similar.

boxed:
Cool.

PR #273, opened 2018-10-23
add freezegun to ignore list to avoid infinite recursion in TickingDa…

sanvyruz:
- Adding "freezegun" into list of ignore packages, to avoid infinite recursion inside TickingDateTimeFactory


boxed:
We should definitely do this, but shouldn't the change just be to add to the default ignore list at https://github.com/spulec/freezegun/blob/master/freezegun/api.py#L718 ?

sanvyruz:
@boxed I did that initially, but then _should_use_real_time returns a True even within the tests causing them to fail. So added it back to default ignore list conditionally when tick=True (which probably works because there is no test with such condition?)

Perhaps a better way would be to able to add "freezegun.api" to ignore list except in testing environment?

PR #281, opened 2019-01-11
Ignore modules with erroneous __dir__ method

johanols:
Modules that overrides its __dir__ method and performing non standard
operations might result in freezegun crashing.

One such example is IPython which adds shim modules when certain
packages are imported. (Eg. `import IPython.html`) This leads to errors
upon activating freezegun if the modules being shimmed are not
installed.

The changes in this commit guards against ImportErrors that might occur
for such modules.

Adding tests to ensure these behaviors are considered.

See: https://github.com/ipython/ipython/blob/5.8.0/IPython/utils/shimmodule.py#L75

### DIFF
--- a/freezegun/api.py
+++ b/freezegun/api.py
@@ -86,7 +86,7 @@ def _get_module_attributes(module):
     result = []
     try:
         module_attributes = dir(module)
-    except TypeError:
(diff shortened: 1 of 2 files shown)


boxed:
Thanks!

PR #288, opened 2019-02-08
Ignore _pytest.runner by default

### LINKED ISSUE #286: Slowest test times from pytest are incorrect
- Python 3.7
- pytest==4.2.0
- freezegun==0.3.11

```python
import pytest
from datetime import timedelta
@pytest.fixture
def freezer():
  with freeze_time(
    # ignore=['_pytest.runner']
  ) as f:
    yield f
def test(freezer):
  freezer.tick(delta=timedelta(seconds=60)
```

```
pytest --durations 1
============= slowest 1 test durations ==============
60.00s call     test_blah.py::test
============= 1 passed in 0.06 seconds ==============
```

If I uncomment the ignore line, the durations match what I expect:

```
============= slowest 1 test durations ==============
0.03s setup    test_blah.py::test
============= 1 passed in 0.08 seconds ==============
```

I also see in the [master branch `_pytest.runner.` was added](https://github.com/spulec/freezegun/blame/master/freezegun/api.py#L729). But this doesn't work for me. Perhaps we should also add `_pytest.runner`?

goodspark:
This will fix pytest duration calculation. Fixes #286

### DIFF
--- a/freezegun/api.py
+++ b/freezegun/api.py
@@ -727,6 +727,7 @@ def freeze_time(time_to_freeze=None, tz_offset=0, ignore=None, tick=False, as_ar
     ignore.append('selenium')
     ignore.append('_pytest.terminal.')
     ignore.append('_pytest.runner.')
+    ignore.append('_pytest.runner')
     
     return _freeze_time(time_to_freeze, tz_offset, ignore, tick, as_arg, auto_tick_seconds)
 
--- a/tests/test_datetimes.py
+++ b/tests/test_datetimes.py
@@ -8,7 +8,7 @@ from unittest import SkipTest
 import pytest
 from tests import utils
 
-from freezegun import freeze_time
+from freezegun import api, freeze_time
 from freezegun.api import FakeDatetime, FakeDate
 
 try:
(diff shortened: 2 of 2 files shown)


freezegun/api.py
```diff
@@ -727,6 +727,7 @@ def freeze_time(time_to_freeze=None, tz_offset=0, ignore=None, tick=False, as_ar
     ignore.append('selenium')
     ignore.append('_pytest.terminal.')
     ignore.append('_pytest.runner.')
+    ignore.append('_pytest.runner')
```
goodspark: `test_should_use_real_time` fails. But if I comment out the lines that change `call_stack_inspection_limit`, they pass.

As a sanity check, I undid my change, left the `call_stack_inspection_limit` lines commented out, and the tests still passed. So I'm not sure what this is doing, other than the comment about increasing coverage.

goodspark: And it looks like the other tests fail because this value never gets reset.

tests/test_datetimes.py
```diff
@@ -628,6 +628,16 @@ def test_time_with_nested():
...
+def change_stack_limit():
+    ori_val = api.call_stack_inspection_limit
+    api.call_stack_inspection_limit = 100  # just to increase coverage
+    yield
+    # Restore to normal after test(s)
+    api.call_stack_inspection_limit = ori_val
+
+
+@pytest.mark.usefixtures('change_stack_limit')
 def test_should_use_real_time():
```
goodspark: `test_should_use_real_time` fails. But if I comment out the lines that change `call_stack_inspection_limit`, they pass.

As a sanity check, I undid my change, left the `call_stack_inspection_limit` lines commented out, and the tests still passed. So I'm not sure what this is doing, other than the comment about increasing coverage.

I think the other tests failed (from the run from the first commit) because the limit was 'leaking' across the tests. The fixture I added would fix this, but perhaps these lines should just be removed?

Issue #268, opened 2018-10-16
Version on PyPI broken for Python 3.3


pganssle:
I believe the latest version of freezegun uploaded on PyPI was uploaded using `python setup.py upload`, which is deprecated. This is a major problem, because you have added `python_requires`, but there is a [bug in distutils/setuptools](https://github.com/pypa/setuptools/issues/1381) that hard-codes the metadata version as 1.0.

This causes the unfortunate situation that now your package will enforce at *install time* that it cannot be installed on Python 3.3, but because PyPI doesn't know anything about that, `pip install freezegun` on Python 3.3 will install the latest version.

See more on this issue [in this warehouse bug](https://github.com/pypa/warehouse/issues/3889).

Here is a [sphinx bug](https://github.com/sphinx-doc/sphinx/issues/4936) where they, unfortunately, decided to do nothing about this.

I would consider un-publishing the latest version and replacing it with a `.post` release uploaded with `twine` instead.

spulec:
We actually dropped support for Python 3.3 here: https://github.com/spulec/freezegun/pull/248

Let me know if there is a good reason for us to reconsider though.

pganssle:
@spulec I know you did, that's the problem.

spulec:
Ah, I see.

The release has already been out for a day. I fear it will cause more damage to try to take it down at this point.

pganssle:
I think if you cut a 0.3.12 release immediately and then unpublish 0.3.11, you'll do almost no damage.

The only people who would be damaged by that are people who have *already* pinned to exactly 0.3.11, as opposed to breaking all Python 3.3 users. Presumably the majority of people pinning to an exact version within a day can also pin to the *next* version within a day.

cpt-jenning:
I would like to draw attention to one case, if you already speak about versions.

Is `auto_tick_seconds` a new develop feature, which should not go to the latest (0.3.11) version? If it's not then project README's misled me. If it should be, then it's absent in PYPI sources.

Great project, I really enjoy using it :)

boxed:
@cpt-jenning  It's not very strange that the github page is ahead of the released code. Development has one pace, and releases have another. There are a lot of features and bugs that are in the backlog that could potentially go into the next release, we'll see what happens.

hugovk:
FreezeGun 0.3.12 was released in May 2019, some months after 0.3.11, so there's probably not much more can be done anymore. 

[...]

PR #296, opened 2019-05-17
Conditionally patch time.clock

pganssle:
`time.clock` was removed in Python 3.8, which is causing `freezegun` to fail when importing the API module as it tries to stash and patch the clock function. This makes the patching logic conditional on the existence of a `time.clock` function.

`time.clock` has been deprecated for some time, and I guess it was just missed in at least my test suite because the *real* function was never invoked.

It may be worth hard failing on deprecation warnings in your test suite in the future, to avoid problems like this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,12 @@ matrix:
     - python: 3.7
       dist: xenial
       sudo: required
+    - python: 3.8-dev
+      dist: xenial
+      sudo: required
+  allow_failures:
+    - python: "3.8-dev"
+
 script: make travis
 install:
   - pip install .
--- a/freezegun/api.py
+++ b/freezegun/api.py
@@ -25,11 +25,13 @@ real_time = time.time
 real_localtime = time.localtime
 real_gmtime = time.gmtime
 real_strftime = time.strftime
-real_clock = time.clock
 real_date = datetime.date
 real_datetime = datetime.datetime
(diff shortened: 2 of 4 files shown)


pganssle:
Test failure on 3.8-dev seems unrelated, but also is emitting deprecation warnings in `time.clock`, so evidently whatever version Travis has doesn't have the `time.clock` removal yet.

boxed:
Nice! Thanks.

pganssle:
@boxed Since this is actively breaking on Python 3.8 *on import*, I think a release sooner rather than later would be a good idea - considering this is a testing library and it may block people from testing 3.8 before the release.

Not sure if you use the `publish` target in the `MakeFile`, but if you do I recommend instead using something like #297 for the publishing instead.

boxed:
Agreed. But I don't have access to make a release. @spulec

spulec:
Thanks!

Released with `0.3.12`.

pganssle:
@spulec It seems like you used the `publish` Makefile target without merging #297, because once again you have hit the problem from #268.  Given that the release was 14 minutes ago, can you please delete it and cut a new one?

Once [PEP 592](https://www.python.org/dev/peps/pep-0592/) is implemented, you will be able to "yank" the last two releases and cut new releases in their place, but you may as well delete the most recent one before someone gets hit by it.

spulec:
Sorry about that. I've merged https://github.com/spulec/freezegun/pull/296 for the future.

PR #297, opened 2019-05-18
Remove setup.py from "publish" target

pganssle:
This replaces `setup.py upload` (which is deprecated and has caused problems for this library before) with `twine upload`, which is the modern way to do it.

I have also gone ahead and set up PEP 517/518 for this project, and used `pep517.build` to build the source and binary distributions, thus removing the need to invoke `setup.py` at all.

This both adds `pyproject.toml` and switches from `setup.py` to `twine`, but if you just want the `twine` command part of it, [the first commit](https://github.com/spulec/freezegun/pull/297/commits/8ccb2a4a8a20ac777e37a975eececb82aafd516b) doesn't include the `pyproject.toml` changes.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,4 @@
-include README.rst LICENSE AUTHORS.rst CHANGELOG
+include README.rst LICENSE AUTHORS.rst CHANGELOG pyproject.toml
 recursive-include tests *
 global-exclude __pycache__
 global-exclude *.py[co]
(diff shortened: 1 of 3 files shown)


pganssle:
I think the reduction in coverage may be an artifact. I have not touched any `.py` files, so presumably `Makefile` or `pyproject.toml` are being erroneously included in the calculation of coverage percentage, and the fact that they have gotten larger makes the "percentage covered" seem like it fell.

boxed:
@pganssle I tend to ignore that check anyway because it's not hugely relevant as you say :)

PR #301, opened 2019-06-05
Update install instructions

The-Compiler:
Freezegun seems to be in Debian Stretch (stable) nowadays: https://packages.debian.org/stretch/python-freezegun


boxed:
Thanks!

Issue #293, opened 2019-03-28
Test failures on Python 3.8


hroncok:
Hi. I'm bootstrapping Python 3.8 in Fedora and I got stuck at freezegun, it has build fialures. Those are reproducible here, on master.

```
$ git rev-parse HEAD
9347d133f33f675c87bb0569d70d9d95abef737f
$ tox -e py38
GLOB sdist-make: .../freezegun/setup.py
py38 inst-nodeps: .../freezegun/.tox/dist/freezegun-0.3.11.zip
py38 installed: atomicwrites==1.3.0,attrs==19.1.0,certifi==2019.3.9,chardet==3.0.4,coverage==4.5.3,coveralls==1.7.0,dateparser==0.7.1,docopt==0.6.2,freezegun==0.3.11,humanize==0.5.1,idna==2.8,maya==0.6.1,mock==2.0.0,more-itertools==6.0.0,pbr==5.1.3,pendulum==2.0.4,pluggy==0.9.0,py==1.8.0,pytest==3.10.1,pytest-cov==2.6.1,pytest-cover==3.0.0,pytest-coverage==0.0,python-dateutil==2.8.0,pytz==2018.9,pytzdata==2018.9,regex==2019.3.12,requests==2.21.0,six==1.12.0,snaptime==0.2.4,tzlocal==1.5.1,urllib3==1.24.1
py38 run-test-pre: PYTHONHASHSEED='905734949'
py38 runtests: commands[0] | make test NOSE_ARGS=
rm -f .coverage
pytest
============================= test session starts ==============================
platform linux -- Python 3.8.0a2, pytest-3.10.1, py-1.8.0, pluggy-0.9.0
rootdir: .../freezegun, inifile:
plugins: cov-2.6.1
collected 105 items

tests/test_asyncio.py ..                                                 [  1%]
tests/test_class_import.py ..............                                [ 15%]
tests/test_datetimes.py ..................................s............. [ 60%]
..........                                                               [ 70%]
tests/test_errors.py ..                                                  [ 72%]
tests/test_import_alias.py ...                                           [ 75%]
tests/test_operations.py .........                                       [ 83%]
tests/test_pickle.py ...                                                 [ 86%]
tests/test_sqlite3.py ..                                                 [ 88%]
tests/test_ticking.py ......                                             [ 94%]
tests/test_utils.py ..                                                   [ 96%]
tests/test_uuid.py FF                                                    [ 98%]
tests/test_warnings.py ..                                                [100%]

[...]

PR #305, opened 2019-06-19
Patch time.time_ns to support Python 3.8

tirkarthi:
* Adds support for patching time.time_ns available from Python 3.7.
* Python 3.8 uses `time.time_ns` in UUID module to fix test case failures.

#293

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ python:
   - 3.5
   - 3.6
   - pypy3
-  - 3.8-dev
 matrix:
   include:
     - python: 3.7
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: python
+cache: pip
 python:
   - 2.7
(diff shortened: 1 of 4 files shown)


tirkarthi:
It would be nice to cache pip to reduce load on PyPI servers. I can do this as part of this PR if okay.

Reference : https://docs.travis-ci.com/user/caching/#pip-cache

boxed:
I'm fine with that. Just say when you're done and I'll merge.

tirkarthi:
I added cache to `.travis.yml` and moved 3.8 from allowed_failures to mandatory to catch any further regressions. GitHub UI is still yellow but the build passed on Travis. Please review the changes.

Thanks

boxed:
Thanks!

tirkarthi:
Thanks for the review. Thanks for the library, we use it at work for datetime related tests and it's very convenient.

PR #308, opened 2019-09-13
Use built mock from unittest when available (Python 3.4+)

kmosiejczuk:
Use the mock built into unittest for Python 3.4+ rather than external mock


spulec:
Thank you!

PR #311, opened 2019-10-13
Add support for calendar.timegm()

### LINKED ISSUE #287: Support for calendar.timegm()
It would be awesome to support [`calendar.timegm()`](https://docs.python.org/3.7/library/calendar.html#calendar.timegm).

It's actually the inverse of `time.gmtime()` which is currently supported and frozen properly.

eric-weaver:
Fixes #287

### DIFF
--- a/freezegun/api.py
+++ b/freezegun/api.py
@@ -26,9 +26,10 @@ real_time = time.time
 real_localtime = time.localtime
 real_gmtime = time.gmtime
 real_strftime = time.strftime
+real_timegm = calendar.timegm
 real_date = datetime.date
 real_datetime = datetime.datetime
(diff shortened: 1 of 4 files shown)


boxed:
Thanks!

ticosax:
We have a regression in our test suite because the value we provide to `calendar.timegm`
gets replaced by the current time

```python
def fake_timegm(struct_time):
    if _should_use_real_time():
        return real_timegm(struct_time)
    else:
        return real_timegm(get_current_time().timetuple())
```

What is the meaning of `_should_use_real_time()` ? which in our case returns False. So the result is unpredictable, because the value we provide gets replaced by current time.
It seems the logic is flawed here.

What do you suggest ?

Issue #314, opened 2019-10-16
Inconsistency between `datetime.now()` and `time.time()`


adaitche:

boxed:
The indent is incorrect in your example I assume?

boxed:
In general freezegun is rather buggy in this respect and people disagree what the behavior should be. I think UTC is the only logical default like you also seem to expect. 

I just haven't had time to work on freezegun for a long time. A good start would be a comprehensive set of tests for what we want the end result to be. Can this example be converted to a unit test?

adaitche:
> The indent is incorrect in your example I assume?

I am not sure what you mean, the indenting seems fine to me. Could you elaborate?

> In general freezegun is rather buggy in this respect and people disagree what the behavior should be. I think UTC is the only logical default like you also seem to expect.

I personally would prefer assuming the local timezone, if the timezone is not explicitly specified. This fits my use case more and seems to me to be more inline with what Python does: it returns datetimes in the local timezone by default. But assuming UTC is also fine.

In any case (whether the datetime passed to freeze_time is assumed to be in UTC or in the local timezone), I think the result of `datetime.now()` and`datetime.fromtimestamp(time.time())` should be the same. If UTC is assumed, then currently the result of `datetime.now()` is not correct as it should return the datetime in the local timezone, i.e. with an offset to the datetime passed to freeze_time.

> I just haven't had time to work on freezegun for a long time. A good start would be a comprehensive set of tests for what we want the end result to be. Can this example be converted to a unit test?

I have submitted a pull request with the tests, which currently fail of course. The second test assumes that the datetime passed to freeze_time is in UTC.

adaitche:
> The local timezone would be super strange. The point of freezing time is to make tests reproducible. Using the local time zone means they are not. Tests could succeed on one persons computer and fail on another because they are in different time zones.

I see your point, but I am not sure how to make it work without introducing inconsistencies. My expectation is that different ways of constructing the current time should have the same results, i.e. it must be that
```python
datetime.now() == datetime.fromtimestamp(time.time())
```
The result of `time.time()` is the unix-time, which is always in UTC. The result of `datetime.now()` in Python is in the local timezone. Making it return the datetime in UTC is not correct IMHO, because it behaves differently when `freeze_time` is used, as compared to the default Python behaviour. Hence, I think `freeze_time` should make `datetime.now()` return the local timezone datetime.

[...]

PR #315, opened 2019-10-17
Test behaviour of datetime.now() in non-UTC timezone

adaitche:
Tests for the inconsistency between `datetime.datetime.now()` and `time.time()`, reported in #314.

### DIFF
--- a/tests/test_datetimes.py
+++ b/tests/test_datetimes.py
@@ -690,3 +690,37 @@ def test_time_ns():
     assert time.time_ns() == expected_timestamp * 1e9
     freezer.stop()
     assert time.time() != expected_timestamp
+
+
+@pytest.mark.skip("timezone handling is currently incorrect")
+def test_compare_datetime_and_time_with_timezone(monkeypatch):
+    """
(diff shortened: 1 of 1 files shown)


tests/test_datetimes.py
```diff
@@ -690,3 +690,34 @@ def test_time_ns():
...
+            dt1 = datetime.datetime.now()
+            dt2 = datetime.datetime.fromtimestamp(time.time())
+            assert dt1 == dt2
+    finally:
+        time.tzset()  # set the timezone back to what is was before
+
+def test_datetime_in_timezone(monkeypatch):
+    """
+    It is assumed that the argument passed to freeze_time is in UTC, unless explicitly indicated
+    otherwise. Therefore datetime.now() should return the fronzen time with an offset.
```
boxed: "fronzen" :P

boxed:
Could you mark these as skipped with a message like "timezone handling is currently incorrect"? We should merge a bunch of tests like this but skipped and then try to fix it...

adaitche:
> Could you mark these as skipped with a message like "timezone handling is currently incorrect"? We should merge a bunch of tests like this but skipped and then try to fix it...

Done.

Issue #303, opened 2019-06-18
Python 3.8 compatibility


JoseKilo:
Using python3.8, with `-Werror`

```python
>>> from freezegun import freeze_time
>>> @freeze_time()
... async def foo():
...     pass
... 
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
  File "[...]/.tox/py38-std-sha256/lib/python3.8/site-packages/freezegun/api.py", line 500, in __call__
    return self.decorate_coroutine(func)
  File "[...]/.tox/py38-std-sha256/lib/python3.8/site-packages/freezegun/api.py", line 689, in decorate_coroutine
    return wrap_coroutine(self, coroutine)
  File "[...]/.tox/py38-std-sha256/lib/python3.8/site-packages/freezegun/_async.py", line 9, in wrap_coroutine
    def wrapper(*args, **kwargs):
  File "/opt/python/3.8-dev/lib/python3.8/asyncio/coroutines.py", line 111, in coroutine
    warnings.warn('"@coroutine" decorator is deprecated since Python 3.8, use "async def" instead',
DeprecationWarning: "@coroutine" decorator is deprecated since Python 3.8, use "async def" instead
```

According to the release notes:
https://docs.python.org/3.8/whatsnew/changelog.html#changelog

> bpo-36921: Deprecate `@coroutine` for sake of async def.

tirkarthi:
One note is that the freezegun supports 3.4 which reached EoL. `async def` syntax is supported only from 3.5 and above

spulec:
Based on some quick research, it doesn't seem like there is an easy way to solve this without dropping support for anything before 3.5. I don't mind dropping support for 3.4, but dropping support for 2.7 is not preferred.

I'm open to any ideas.

eric-weaver:
I put up #316 as an attempt to fix this issue. Let me know if this makes sense

spulec:
This should be fixed with https://github.com/spulec/freezegun/pull/316. Let me know if you see other issues.

FedericoCeratto:
Related to https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=944026

PR #316, opened 2019-10-18
Use async def for python > 3.5

eric-weaver:
Took a stab at fixing #303. Modified `_async.py` to make use of `async def` keywords and moved the old `@asyncio.coroutine` style into a new `_async_coroutine.py` file. Conditionally import `_async_coroutine.py` only on python < 3.5. I purposely left the usage of `@asyncio.coroutine` in [test_asyncio.py](https://github.com/spulec/freezegun/blob/011138dc548bcf77c237a9f4f334b2838aa9d10b/tests/test_asyncio.py#L17) as the tests will fail due to syntax errors when run against older versions of python. This should fix the current issue when running under `python3.8 -Werror`.

@spulec  let me know if this looks like a decent direction or if I am missing something else around the original issue.

### DIFF
--- a/freezegun/_async_coroutine.py
+++ b/freezegun/_async_coroutine.py
@@ -5,12 +5,13 @@ import asyncio
 
 def wrap_coroutine(api, coroutine):
     @functools.wraps(coroutine)
-    async def wrapper(*args, **kwargs):
+    @asyncio.coroutine
+    def wrapper(*args, **kwargs):
         with api as time_factory:
             if api.as_arg:
-                result = await coroutine(time_factory, *args, **kwargs)
+                result = yield from coroutine(time_factory, *args, **kwargs)
             else:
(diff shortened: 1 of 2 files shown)


spulec:
Looks great; thanks!

Keruspe:
Fwiw, this broke python 2.7 support (not sure if we care though)

```
Compiling /var/tmp/paludis/build/dev-python-freezegun-0.3.13/image//usr/x86_64-pc-linux-gnu/lib/python2.7/site-packages/freezegun/_async_coroutine.py ...      
  File "/usr/x86_64-pc-linux-gnu/lib/python2.7/site-packages/freezegun/_async_coroutine.py", line 12                                                           
    result = yield from coroutine(time_factory, *args, **kwargs)
                      ^                    
SyntaxError: invalid syntax      
```

PR #317, opened 2019-10-18
Make string_type private to fix generated type stubs

ska-kialo:
While generating type stubs for freezegun for usage with mypy I noticed that the output is broken. Instead of generating just one assignment of `string_type` it generates both, leading to errors in mypy when checking python 3 code since `basestring` doesn't exist.

I didn't find an easy way to fix this in mypy so I thought the simplest solution was to make the `string_type` in freezegun private since mypy's stubgen tool doesn't generate definitions for private names.

Stubgen output before:
```python
string_type = basestring
string_type = str
```

After: nothing, since the type alias is now private.

### DIFF
--- a/freezegun/api.py
+++ b/freezegun/api.py
@@ -48,9 +48,9 @@ tick_flags = []
 # Python3 doesn't have basestring, but it does have str.
 try:
     # noinspection PyUnresolvedReferences
-    string_type = basestring
+    _string_type = basestring
 except NameError:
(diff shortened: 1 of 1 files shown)


boxed:
I've also tried to change mypy and given up. This seems like an ok temporary workaround until we drop py2 support.

ska-kialo:
Awesome, thanks for merging this so quickly 👍

Issue #318, opened 2019-10-28
Add mypy types (or type stubs)


ska-kialo:
Hi,

my colleague(s) and I are currently working on writing (a subset of) type stubs for freezegun and we were wondering if it would make sense to contribute them.

Going by the [contribution guide for typeshed](https://github.com/python/typeshed/blob/master/CONTRIBUTING.md#adding-a-new-library) the first step in the process is asking the maintainers of the libraries if they want to integrate the type hints into the library directly.
What is your opinion on this?

Contributing the types here could be done one of the following ways:

- adding the types as separate stub files
- adding the types in the code as annotations (this isn't compatible with Python 2)
- adding the types as type comments (compatible with Python 2 and 3)

boxed:
I think adding them as annotations is the way to go. We are going to drop python 2 support for new versions at year end anyway. Or at least I think we should :P

@spulec what do you think?

boxed:
I'm fine with this too. People who are on python 3 are mostly on 3.5+ anyway and if they need it for older versions they can use older versions of freezegun.

spulec:
Works for me

sbdchd:
Type stubs were recently added to the type shed, although it's preferable to keep them in the project itself

https://github.com/python/typeshed/pull/4425

PR #319, opened 2019-10-29
Add type annotations

ska-kialo:
Concerns #318.

A first shot of adding type annotations to the library.
This is just for getting some feedback there might be some clean up work to do.

A few notes:
* As discussed in #318, this would drop support for any Python version below 3.5 to allow using type annotations instead of type comments. This also simplifies things like string types, `copyreg` and coroutine handling.
* Coroutine handling could be further simplified using only the `async def` variant seen in #316 (since that is supported since 3.5)
* Type comments (`# type: ...`) are used for variables and class fields since annotating these directly would require dropping support for Python 3.5 (EOL is 09/2020).
* There are quite a few `# type: ignore` comments that are (as far as I can tell) unavoidable because they deal with private fields from stdlib modules and assigning to types (which mypy doesn't allow)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,6 +28,8 @@ setup(
     author_email='spulec@gmail.com',
     url='https://github.com/spulec/freezegun',
     packages=['freezegun'],
+    package_data={'freezegun':['py.typed']},
+    zip_safe=False,
     install_requires=requires,
     tests_require=tests_require,
     include_package_data=True,
--- a/setup.py
+++ b/setup.py
@@ -3,18 +3,8 @@
 import sys
 from setuptools import setup
 
-requires = ['six']
-tests_require = [
-    'mock;python_version<"3.4"',
-    'nose'
-]
-
-if sys.version_info.major == 2:
-    requires += ['python-dateutil>=1.0, != 2.0']
-else:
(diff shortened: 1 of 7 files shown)


setup.py
```diff
@@ -3,18 +3,8 @@
...
-    'nose'
-]
-
-if sys.version_info.major == 2:
-    requires += ['python-dateutil>=1.0, != 2.0']
-else:
-    # Py3k
-    requires += ['python-dateutil>=2.0']
-
+requires = ['six', 'python-dateutil>=2.0']
```
syurchenko: do we still need six library if we drop py2 support?

ska-kialo: You're right, we can remove that dependency then

staticdev:
@spulec @boxed with the latest mypy, freezegun now have very annoying errors for everyone using it. Should it possible to go forward with type-hinting into this package? Adding another package just to get freezegun is very overengineering.

boxed:
This PR is marked as work in progress. I don't know why.

gitpushdashf:
I'm interested in this as well.

staticdev:
@gitpushdashf I created a new PR for that #403

Issue #89, opened 2015-03-09
Faked time shouldn't always be the same.


homeworkprod:
Excerpt from the docs:
"[…] all calls to ``datetime.datetime.now()``, ``datetime.datetime.utcnow()``, […] will return the time that has been frozen."

But this is not the behaviour I expect.

Example code:
```python
from datetime import datetime

def test_something():
    print('now():   ', datetime.now())
    print('utcnow():', datetime.utcnow())
```

Output without Freezegun (system timezone is "Europe/Berlin"):
```
now():    2015-03-10 00:46:40.465879
utcnow(): 2015-03-09 23:46:40.466031
```

Output with Freezegun (method decorated with ``@freeze_time('2012-03-11 19:21:23')``):
```
now():    2012-03-11 19:21:23
utcnow(): 2012-03-11 19:21:23
```

However, I expected the output with Freezegun to reflect the offset (one hour, in this case) and, thus, to return two different date/time instances.

homeworkprod:
Oh, and Freezegun's fake date/time lacks milliseconds compared to the default update. I've tested on Python 3.4.2.

spulec:
Thank you for opening this. Your point is very valid. I would like to fix this somehow, but have a few thoughts/questions I'm trying to think through. Any feedback appreciated.

- I think we would need to have some sort of backwards-compatibility. No idea how this would work.
- Should the time passed into freezegun be used as the local time or utc time?
- What if the time passed in has a timezone as well or a `tz_offset` is passed?

I'm also quite worried about how the documentation will read.

homeworkprod:
> I think we would need to have some sort of backwards-compatibility. No idea how this would work.
Staying completely backwards-compatible probably won't work.

> Should the time passed into freezegun be used as the local time or utc time?
I'd say this should be the local time. If it should be UTC, either a separate function with some `_utc` suffix should be provided (just like `datetime.now()` and `datetime.utcnow()` in the stdlib), or a keyword argument like `utc=True` or similar could be used.

> What if the time passed in has a timezone as well or a tz_offset is passed?
Passing an optional timezone via a keywords argument should be possible. If that is the case and the time value already specifies timezone, a `ValueError` could be raised.

agriffis:
I agree it should be the local time. Caller can specify UTC as the timezone if they want that.

agriffis:
Actually, it should probably default to UTC, so that developers in different timezones working on the same project won't see different results in their tests.

[...]

PR #323, opened 2019-12-17
Update unit tests to pass in different timezones

### LINKED ISSUE #322: Unit tests fail when run in Adelaide, Australia
Hi, I've been having a play with your nice library but I've discovered that lots of unit tests fail when my machine is in my local timezone "Australian Central Daylight Time". The unit tests pass if I change to a northern hemisphere timezone like Central Standard Time.

I've inspected one of the tests and it appears to assume that it is not DST during 2012-01-14 which is true for the North America, but not Australia.

I'm happy to have a go at fixing this, and I have a fix that I think works, but would like your feedback on if it is the correct way to go.

dcrowe:
Fixes #322

### DIFF
--- a/tests/test_datetimes.py
+++ b/tests/test_datetimes.py
@@ -51,9 +51,8 @@ def test_simple_api():
     # time to freeze is always provided in UTC
     freezer = freeze_time("2012-01-14")
     # expected timestamp must be a timestamp, corresponding to 2012-01-14 UTC
-    local_time = datetime.datetime(2012, 1, 14)
-    utc_time = local_time - datetime.timedelta(seconds=time.timezone)
-    expected_timestamp = time.mktime(utc_time.timetuple())
+    utc_time = datetime.datetime(2012, 1, 14, tzinfo=datetime.timezone.utc)
+    expected_timestamp = calendar.timegm(utc_time.timetuple())
 
     freezer.start()
     assert time.time() == expected_timestamp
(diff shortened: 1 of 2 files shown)


dcrowe:
That _Utc class was totalled not needed. It is much simpler now.

I understand that this doesn't interact well with #89, but I'm looking for a smaller issue to chew on for my first contribution. :)

Tested in the following timezones:
```
# not in DST
TZ=Europe/Berlin make test
# in DST
TZ=Australia/Adelaide make test
# doesn't have DST
TZ=Australia/Queensland make test
# UTC
TZ=UTC make test
```

tests/test_datetimes.py
```diff
@@ -51,9 +51,8 @@ def test_simple_api():
     # time to freeze is always provided in UTC
     freezer = freeze_time("2012-01-14")
     # expected timestamp must be a timestamp, corresponding to 2012-01-14 UTC
-    local_time = datetime.datetime(2012, 1, 14)
-    utc_time = local_time - datetime.timedelta(seconds=time.timezone)
-    expected_timestamp = time.mktime(utc_time.timetuple())
+    utc_time = datetime.datetime(2012, 1, 14)
+    expected_timestamp = calendar.timegm(utc_time.timetuple())
```
pganssle, line 55: This seems like a sorta clunky way to get a timestamp for UTC, how about this:

```python
utc_time = datetime.datetime(2012, 1, 14, tzinfo=datetime.timezone.utc)
expected_timestamp = utc_time.timestamp()
```

tests/test_datetimes.py
```diff
@@ -456,9 +454,8 @@ def test_nested_context_manager():
 def _assert_datetime_date_and_time_are_all_equal(expected_datetime):
     assert datetime.datetime.now() == expected_datetime
     assert datetime.date.today() == expected_datetime.date()
-    datetime_from_time = datetime.datetime.fromtimestamp(time.time())
-    timezone_adjusted_datetime = datetime_from_time + datetime.timedelta(seconds=time.timezone)
-    assert timezone_adjusted_datetime == expected_datetime
+    datetime_from_time = datetime.datetime.utcfromtimestamp(time.time())
```
pganssle, line 457: I'm not convinced that this is the right thing to do, but I also don't think the assertion this replaces makes much sense.

It seems to be asserting that `datetime.datetime.now()` should *not* be equal to `datetime.fromtimestamp(time.time())` if the test is run in a locale with an offset, which is not the same behavior as what `datetime.datetime` does. Fixing this assumption in the tests (and code?) is how this should be fixed, I think.

pganssle:
I think it would probably be a good idea to add a CI job that runs the tests in some non-UTC locale (or possibly in a few locales). Presumably it's OK to just use whatever the latest Python is and run the test suite in a `America/New_York`, `Australia/Adelaide` and `Europe/London` (or even just one of those, to start).

PR #328, opened 2020-01-20
Fix locale timestamp

### LINKED ISSUE #208: datetime.datetime.fromtimestamp returns unexpected values
`datetime.datetime.fromtimestamp` [should return](https://docs.python.org/2/library/datetime.html#datetime.date.fromtimestamp) a local date, but the system timezone is used and the `tz_offset` parameter seems ignored.

```python
import freezegun, datetime, time
with freezegun.freeze_time("1970-01-01 00:00:01"):
    print(datetime.datetime.fromtimestamp(0))

with freezegun.freeze_time("1970-01-01 00:00:01", tz_offset=6):
    print(datetime.datetime.fromtimestamp(0))
```
```bash
# freezegun behavior
$ env TZ=UTC-12 python fromtimestamp.py
1970-01-01 12:00:00
1970-01-01 12:00:00

# Expected results
1970-01-01 00:00:00
1970-01-01 06:00:00
```

isra17:
Should fix #208 and #314 

`timestamp` and `fromtimestamp` relied on locale timezone to calculate their values. This fix ensure `tz_offset` is used instead, fixing a few API inconsistencies as seen in the tests.

### DIFF
--- a/freezegun/api.py
+++ b/freezegun/api.py
@@ -21,6 +21,8 @@ except ImportError:
     MayaDT = None
 
 _TIME_NS_PRESENT = hasattr(time, 'time_ns')
+_EPOCH = datetime.datetime(1970, 1, 1)
+_EPOCHTZ = datetime.datetime(1970, 1, 1, tzinfo=datetime.timezone.utc)
 
 real_time = time.time
 real_localtime = time.localtime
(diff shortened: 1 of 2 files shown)


boxed:
Thanks!

luizribeiro:
@spulec, are you planning on making a release that will contain this fix anytime soon? We've been facing the issue from #314 in some tests.

spulec:
Published as `0.3.15`

PR #342, opened 2020-03-24
Fix bug in tzinfo

### LINKED ISSUE #341: dateutil.tz has no attribute UTC
I've encountered this issue after installing freezegun in my project and try to run in my tests:

```
    _EPOCHTZ = datetime.datetime(1970, 1, 1, tzinfo=dateutil.tz.UTC)
AttributeError: module 'dateutil.tz' has no attribute 'UTC'

```
All the dependencies is update with the last modification.

vinimmelo:
Fix a bug in freezegun datetime tzinfo parameter.

closes #341

### DIFF
--- a/freezegun/api.py
+++ b/freezegun/api.py
@@ -23,7 +23,7 @@ except ImportError:
 
 _TIME_NS_PRESENT = hasattr(time, 'time_ns')
 _EPOCH = datetime.datetime(1970, 1, 1)
-_EPOCHTZ = datetime.datetime(1970, 1, 1, tzinfo=dateutil.tz.UTC)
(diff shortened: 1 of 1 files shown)


boxed:
According to the code https://github.com/dateutil/dateutil/blob/master/dateutil/tz/tz.py#L128 it was added in 2.7.0 in march 2018. Could you verify you are indeed using such an old version of dateutil?

vinimmelo:
Yes, you are absolutely right!!! My python-dateutil version was 2.6.1 and not the latest. Thank you and sorry for the unecessary PR. :wink:

PR #353, opened 2020-05-28
Added as_kwarg argument

### LINKED ISSUE #262: as_arg adds frozen_time argument to front of list of arguments
Currently the following code does not work with test classes (Django, etc). 

https://github.com/spulec/freezegun/blob/master/freezegun/api.py#L669

That line does the following
```python
result = func(time_factory, *args, **kwargs)
```

That is incompatible with `self` being the first param to class methods and it behaves differently than most other decorators, which add their argument to the end, not the beginning.  Here is an example that doesn't work

```python
from django.test import TestCase


class MyTest(TestCase):

  @freeze_time('2016-11-01', as_arg=True)
  def test_something_with_time(self, time_factory):
    # freeze_time overrides self instead of time_factory
    # self is now the time_factory variable
    print(self) # <TimeFactory>
    print(time_factory) # <MyTestCase>
```

It should do this instead.

```python
result = func(*args, time_factory, **kwargs)
```

This is a backwards incompatible change, but will allow the decorator to be used with class based tests. Would you be open to accepting a PR that makes this breaking change?

WisdomPill:
Closes #262 

Added as_kwarg argument in order to have the frozen time object passed with the name provided in as_kwarg argument. Like the following

``` python
class TestUnitTestMethodDecorator(unittest.TestCase):
    @freeze_time('2013-04-09')
    def test_method_decorator_works_on_unittest(self):
        self.assertEqual(datetime.date(2013, 4, 9), datetime.date.today())

    @freeze_time('2013-04-09', as_kwarg='frozen_time')
    def test_method_decorator_works_on_unittest(self, frozen_time):
        self.assertEqual(datetime.date(2013, 4, 9), datetime.date.today())
        self.assertEqual(datetime.date(2013, 4, 9), frozen_time.time_to_freeze.today())

    @freeze_time('2013-04-09', as_kwarg='hello')
    def test_method_decorator_works_on_unittest(self, **kwargs):
        self.assertEqual(datetime.date(2013, 4, 9), datetime.date.today())
        self.assertEqual(datetime.date(2013, 4, 9), kwargs.get('hello').time_to_freeze.today())
```

### DIFF
--- a/freezegun/api.py
+++ b/freezegun/api.py
@@ -730,7 +730,7 @@ class _freeze_time(object):
         def wrapper(*args, **kwargs):
             with self as time_factory:
                 if self.as_arg and self.as_kwarg:
-                    raise Exception()
+                    assert False, "You can't specify both as_arg and as_kwarg at the same time. Pick one."
                 elif self.as_arg:
                     result = func(time_factory, *args, **kwargs)
                 elif self.as_kwarg:
--- a/freezegun/api.py
+++ b/freezegun/api.py
@@ -513,8 +513,7 @@ class StepTickTimeFactory(object):
 
 class _freeze_time(object):
 
-
-    def __init__(self, time_to_freeze_str, tz_offset, ignore, tick, as_arg, auto_tick_seconds):
+    def __init__(self, time_to_freeze_str, tz_offset, ignore, tick, as_arg, as_kwarg, auto_tick_seconds):
         self.time_to_freeze = _parse_time_to_freeze(time_to_freeze_str)
         self.tz_offset = _parse_tz_offset(tz_offset)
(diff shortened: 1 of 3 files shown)


freezegun/api.py
```diff
@@ -729,8 +729,13 @@ def decorate_coroutine(self, coroutine):
     def decorate_callable(self, func):
         def wrapper(*args, **kwargs):
             with self as time_factory:
-                if self.as_arg:
+                if self.as_arg and self.as_kwarg:
+                    raise Exception()
```
boxed, line 733: It's better to do `assert False, 'You can't specify both as_arg and as_kwarg'`

boxed:
Thanks!

PR #356, opened 2020-06-27
Fix inconsistency between timestamp and strftime

### LINKED ISSUE #343: Inconsistency betewen timestamp() and strftime('%s')
Consider the example below:
```
import freezegun
from datetime import datetime

with freezegun.freeze_time("1970-01-01 00:00"):
    print(datetime(2010, 10, 10).timestamp());
    print(datetime(2010, 10, 10).strftime('%s'))
```
it prints different values for each of the methods:
```
1286668800.0
1286661600
```

Results are the same for the native datetime object

```
In [7]: datetime(2010, 10, 10).timestamp()
Out[7]: 1286661600.0
```

Seems something is wrong with the way `timestamp()` works on `FakeDatetime`

mkarys:
It fixes #343

### DIFF
--- a/freezegun/api.py
+++ b/freezegun/api.py
@@ -219,6 +219,10 @@ def fake_strftime(format, time_to_format=None):
     if time_to_format is None:
         return real_strftime(format)
     else:
+        y, m, d, hh, mm, ss = time_to_format[:6]
+        shifted_time = datetime.datetime(y, m, d, hh, mm, ss) \
+                       - datetime.timedelta(seconds=time.timezone)
+        time_to_format = shifted_time.utctimetuple()
         return real_strftime(format, time_to_format)
 
 if real_clock is not None:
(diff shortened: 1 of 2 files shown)


boxed:
I'm confused about the test... doesn't this test basically do nothing if run on a system with UTC as local time zone?

The test is very elegant, but it seems like it relies on a misfeature of freezegun.. or am I misunderstanding?

mkarys:
@boxed yes, you right, it fixes cases that UTC is not local time zone, if UTC is local time zone it was okay without it.
What misfeature of freezegun are you refer to?

pganssle:
I don't think this is the right solution. It fixes the problem with `"%s"` but breaks `strftime` in general.

For example, running this with `TZ='America/New_York'`:

```python
import freezegun
from datetime import datetime

with freezegun.freeze_time("1970-01-01T00:00"):
    print(datetime(2010, 1, 1).strftime("%Y-%m-%dT%H:%M:%S"))
    # 2009-12-31T19:00:00

print(datetime(2010, 1, 1).strftime("%Y-%m-%dT%H:%M:%S"))
# 2010-01-01T00:00:00
```

PR #362, opened 2020-07-14
Travis CI: Test on Python 3.9

hugovk:
Python 3.9 is in beta with final 3.9.0 release due in October.

* https://www.python.org/dev/peps/pep-0596/

Test now on 3.9-dev to prevent surprises at release.


hugovk:
Updated now 3.9 final is out.

boxed:
Thanks!

PR #363, opened 2020-07-17
[cleanup] Fix CI by removing maya for pypy3 test and some various cleanup

Djailla:


boxed:
I don't think we should drop pypy...

Djailla:
It fails for testing, not to deploy.

About CI, any though migrating to github actions ?

boxed:
This commit does remove pypy from setup.py and I don't trust support if we can't get it to work in CI. This isn't good imo.

boxed:
As for github actions: I don't really care. I have more experience with Travis.

Djailla:
Hello @boxed I did some changes to the PR if you want to check

boxed:
Thanks for the fixes!

PR #364, opened 2020-08-13
Hack for github/markup rendering SVG image badges

mondeja:
Github does not display well badges with SVG images as RST references. I've [opened a pull](https://github.com/github/markup/pull/1368) to solve this bug, but the project does not seems maintained.

You can note that hovering the PyPI version badge appends an underline after it like a link. Adding `#foo` after the link solves this.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,7 +1,7 @@
 FreezeGun: Let your Python tests travel through time
 ====================================================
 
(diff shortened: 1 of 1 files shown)


boxed:
Is the little underscore the entire problem?

mondeja:
> Is the little underscore the entire problem?

Yes.

PR #370, opened 2020-10-07
Allow to configure default ignore list

lukaszb:
Backstory: some tests started to fail with some cryptic errors. Ended up it started happening after adding `tenserflow` package. Now we need to somehow remember to always ignore it or create wrapper around `freeze_time` calls.

With ability to configure default ignore list it would be much simpler.

Let me know if this makes sense.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -17,3 +17,4 @@ Patches and Suggestions
 - `Dan Elkis <github.com/rinslow>`_
 - `Bastien Vallet <github.com/djailla>`_
 - `Julian Mehnle <github.com/jmehnle>`_
+- `Lukasz Balcerzak <https://github.com/lukaszb>`_
--- a/README.rst
+++ b/README.rst
@@ -264,3 +264,48 @@ On Debian systems:
 .. code-block:: bash
 
     $ sudo apt-get install python-freezegun
+
+
+Ignore packages
+---------------
+
+Sometimes it's desired to ignore FreezeGun behaviour for particular packages (i.e. libraries).
+It's possible to ignore them for a single invocation:
+
+
+.. code-block:: python
+
+    from freezegun import freeze_time
+
+    with freeze_time('2020-10-06', ignore=['threading']):
+        # ...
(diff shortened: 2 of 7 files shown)


boxed:
I like it!

How about changing

```python
def configure(default_ignore=None):
    settings.default_ignore = default_ignore
```

to something like:


```python
def configure(default_ignore=None, add_ignores=None):
    assert default_ignore is None or add_ignores is None, "You can only override the full list, or add to the default. Both at the same time doesn't make sense"
    if default_ignore is not None:
        settings.default_ignore = default_ignore
    else:
        settings.default_ignore = DEFAULT_IGNORE + add_ignores
```

Writing this I think some renames would be nice too:

DEFAULT_IGNORE -> DEFAULT_IGNORE_LIST

default_ignore -> default_ignore_list
add_ignores -> add_to_ignore_list

lukaszb:
@boxed makes sense, done. I've swapped `assert` to custom exception as assertions might be omitted (i.e. if run with `python -O`)

boxed:
Nice.

I don't agree with the custom exception though. Assertions can be omitted sure, but that's pretty pathological since this library is primary for tests and running tests with asserts turned off defeats the purpose :)

lukaszb:
Right, good point :) Still I believe custom exception is more elegant. But that's my opinion.

Anyway, wanna me to change exception to assertion? Any other changes or it looks like something that can be accepted?

boxed:
Yea, let's not argue over irrelevant details. MERGE! :)

remik:
Hi @boxed 
Could you release it?

boxed:
I can't unfortunately, I don't have access on pypi. @spulec must either give me permission or make a release.

spulec:
@boxed sorry for my recent absence. I've added you on PyPI. Feel free to make a release as helpful.

remik:
Great so @boxed all for you :)

boxed:
Published a new release!

remik:
Great. Thank you!

Issue #345, opened 2020-04-10
"IndexError: list index out of range" when accessing ignore_lists or freeze_factories from thread


alex-kowalczyk:
When fake_time or similar patched thing is somehow passed (eg via queue) to some thread, it might be later called there.
In the meanwhile the patching might have been removed (e.g main thread exited `with freeze_time()` block). This resuls in ignore_lists or freeze_factories being empty.

This bugs hits logging background transports heavily. 
 
Quick and dirty solution would be to check for presence of the element [-1] before accessing it.

This results in an exception like these:
`  File "/opt/python3.7/lib/python3.7/threading.py", line 926, in _bootstrap_inner
    self.run()
  File "/opt/python3.7/lib/python3.7/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/env/lib/python3.7/site-packages/google/cloud/logging/handlers/transports/background_thread.py", line 141, in _thread_main
    max_latency=self._max_latency,
  File "/env/lib/python3.7/site-packages/google/cloud/logging/handlers/transports/background_thread.py", line 67, in _get_many
    elapsed = time.time() - start
  File "/env/lib/python3.7/site-packages/freezegun/api.py", line 173, in fake_time
    if _should_use_real_time():
  File "/env/lib/python3.7/site-packages/freezegun/api.py", line 151, in _should_use_real_time
    if not ignore_lists[-1]:
IndexError: list index out of range" `

Or: 
` File "/usr/lib/python3.7/socketserver.py", line 650, in process_request_thread
    self.finish_request(request, client_address)
  File "/usr/lib/python3.7/socketserver.py", line 360, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/usr/lib/python3.7/socketserver.py", line 720, in __init__
    self.handle()
  File "site-packages/django/core/servers/basehttp.py", line 155, in handle
    handler.run(self.server.get_app())
  File "/usr/lib/python3.7/wsgiref/handlers.py", line 148, in run
    self.close()
  File "/usr/lib/python3.7/wsgiref/simple_server.py", line 35, in close
    self.status.split(' ',1)[0], self.bytes_sent
  File "/usr/lib/python3.7/http/server.py", line 544, in log_request
    self.requestline, str(code), str(size))
  File "site-packages/django/core/servers/basehttp.py", line 99, in log_message
    'server_time': self.log_date_time_string(),
[...]

PR #374, opened 2020-11-02
Fix freezegun when accessing from thread after stop()

juliomalegria:
There is more context on issue #345.

This branch checks if `ignore_lists` is empty before checking the last element. If it is empty, means `close()` has already been called, and we should return the real time.

This only happens when a thread is created, since `ignore_lists` is a global variable.

### DIFF
--- a/freezegun/api.py
+++ b/freezegun/api.py
@@ -167,6 +167,10 @@ def _should_use_real_time():
     if not call_stack_inspection_limit:
         return False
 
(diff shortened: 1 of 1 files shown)


boxed:
Thanks!

juliomalegria:
Wow! Thanks for the swift merge. Any idea when there'll be a new release to be able to use get this from `pip`?

boxed:
Unfortunately I have no power over that.

@spulec ?

Issue #384, opened 2021-01-24
Freezing monotonic time implicitly causes problems


tzoiker:
@boxed I have faced the very same problem as in #383, but in another context and it was very painful to debug.

- I have used `aiormq` library to connect to RabbitMQ;
- It uses heartbeats to monitor connections and relies on monotonic time to set the time of the last heartbeat;
- Using `freeze_time` in some tests lead to the following:
    - Last heartbeat was set to event loop time;
    - `freeze_time` patched monotonic time;
    - Difference between the last heartbeat and the loop time became extremely large, leading to connection errors;
The problem was aggravated by the fact that I didn't pin the minor version of `freezegun` in the testing pipeline so that when `1.1.0` was released tests started to fail unexpectedly.

Some suggestions:
- Do not ever patch monotonic time **implicitly** for any module - it will most certainly break things. Adding asyncio to default ignore list will not help (there may be different event loop implementations) in all the cases which rely on continuous flow of monotonic time.
- Instead of ignoring specific modules, it should be possible to specify exactly what kind of "times" user wants to freeze, and once again, monotonic time should not be there by default.

Thank you for consideration.

shughes-uk:
This also causes issues with celery works due to kombu relying on monotonic https://github.com/celery/kombu/blob/master/kombu/transport/virtual/base.py#L950-L965

ExpLog:
This causes all manner of problems with `aiokafka` as well, which are nearly impossible to debug, as it apparently just hangs when you give it timeouts.

Be very careful what you wrap with `freezetime` if you're using async.

My solution for now was to wrap factory fixtures with `freezetime` instead of wrapping tests.

PR #388, opened 2021-03-04
Support selective patching

tzoiker:
Motivated by #384 - implicit monotonic time patching may (and does) cause hard-to-debug problems in various settings.

- Add `Target` enum describing functions that can be patched;
- Pass `targets` parameter to `freeze_time` to describe functions to be patched;
- Make `targets` exclude monotonic time by default;
- Add `freeze_time_with_monotonic` helper to include monotonic time patching.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -13,7 +13,7 @@ FreezeGun is a library that allows your Python tests to travel through time by m
 Usage
 -----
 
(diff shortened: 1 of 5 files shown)


Pavkazzz:
@spulec Can we merge this one? 
Looks good

PR #395, opened 2021-05-01
Add support for perf_counter

pganssle:
`time.perf_counter` and `time.monotonic` are similar clocks (on many platforms they are actually the *same* clock), with the main difference between them being that if a system has a higher resolution monotonic clock than the one that `time.monotonic()` calls, it will be used for `time.perf_counter()`.

This commit adds support for `time.perf_counter()` and `time.perf_counter_ns()` and refactors `monotonic()` and `monotonic_ns()` so that these similar functions can share an implementation.

It also refactors some of the tests around `monotonic()` and `monotonic_ns()` so that the same tests can be used for `perf_counter(_ns)` as well.

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -3,6 +3,8 @@ from unittest import SkipTest
 
 from freezegun.api import FakeDate, FakeDatetime, _is_cpython
 
+import pytest
+
 
 def is_fake_date(obj):
     return obj.__class__ is FakeDate
@@ -10,6 +12,11 @@ def is_fake_date(obj):
 
 def is_fake_datetime(obj):
     return obj.__class__ is FakeDatetime
+
+
+cpython_only_mark = pytest.mark.skipif(
+    not _is_cpython,
+    reason="Requires CPython")
 
 
 def cpython_only(func):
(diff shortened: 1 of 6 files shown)


tests/utils.py
```diff
@@ -12,6 +14,11 @@ def is_fake_datetime(obj):
     return obj.__class__ is FakeDatetime
 
 
+cpython_only_mark = pytest.mark.skipif(
```
pganssle, line 17: The `cpython_only` implementation is incompatible with `pytest.mark.parametrize` as far as I can tell, and so I had to use a `pytest` mark instead.

I think that `cpython_only` can be redefined to be this mark, but doing so breaks the `test_utils` test for `cpython_only`, which relies on a certain implementation detail. Personally, I'd rather drop the test for `cpython_only` (rely on coverage metrics to make sure your tests aren't being skipped erroneously) and redefine it as a `pytest` mark, but I wanted to keep the scope of this PR small.

pganssle:
Hoping to ping on this for any maintainers watching. Of course, asking for a review for my PR triggers my guilt reflex, so I reviewed a few other PRs in the queue before doing so, to hopefully balance things out. ☺

boxed:
LGTM. Thanks!

PR #397, opened 2021-05-07
Fix decorate_class for Python 3.10 where staticmethod is callable.

### LINKED ISSUE #396: Python 3.10: TypeError: Tester.helper() takes 0 positional arguments but 1 was given
When running tests on Python 3.10.0b1, I get the following test failure:

```
_______________________________________________ Tester.test_class_decorator_respects_staticmethod ________________________________________________

self = <tests.test_datetimes.Tester object at 0x7fb072c03be0>

    def test_class_decorator_respects_staticmethod(self):
>       assert self.helper() == datetime.date(2012, 1, 14)

tests/test_datetimes.py:450: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

args = (<tests.test_datetimes.Tester object at 0x7fb072c03be0>,), kwargs = {}
time_factory = <freezegun.api.FrozenDateTimeFactory object at 0x7fb072c01810>

    def wrapper(*args, **kwargs):
        with self as time_factory:
            if self.as_arg and self.as_kwarg:
                assert False, "You can't specify both as_arg and as_kwarg at the same time. Pick one."
            elif self.as_arg:
                result = func(time_factory, *args, **kwargs)
            elif self.as_kwarg:
                kwargs[self.as_kwarg] = time_factory
                result = func(*args, **kwargs)
            else:
>               result = func(*args, **kwargs)
E               TypeError: Tester.helper() takes 0 positional arguments but 1 was given

freezegun/api.py:778: TypeError
```

tirkarthi:
Earlier staticmethod was not callable and hence decoration was skipped. Now with Python 3.10 they are callable but detect the staticmethod behavior better so that they represent the behavior before Python 3.10.

Fixes #396

### DIFF
--- a/freezegun/api.py
+++ b/freezegun/api.py
@@ -598,7 +598,10 @@ class _freeze_time(object):
                         continue
                     seen.add(attr)
 
-                    if not callable(attr_value) or inspect.isclass(attr_value):
+                    # staticmethods are callable from Python 3.10 . Hence skip them from decoration
+                    if (not callable(attr_value)
+                        or inspect.isclass(attr_value)
+                        or isinstance(attr_value, staticmethod)):
                         continue
 
                     try:


freezegun/api.py
```diff
@@ -598,7 +598,10 @@ def tearDownClass(cls):
                         continue
                     seen.add(attr)
 
-                    if not callable(attr_value) or inspect.isclass(attr_value):
+                    # staticmethods are callable from Python 3.10 . Hence skip them from decoration
+                    if (not callable(attr_value)
+                        or inspect.isclass(attr_value)
+                        or isinstance(attr_value, staticmethod)):
```
mgorny, line 604:
```suggestion
                            or inspect.isclass(attr_value)
                            or isinstance(attr_value, staticmethod)):
```

I don't think you're really supposed to use the same indent level as the console inside `if`.

PR #406, opened 2021-07-23
Fix now(tz)

mkenigs:
now(tz) should return the same time utcnow returns adjusted by whatever
offset is contained by tz. Currently, the offset to freeze_time is also
added, which is removed by this change


emontnemery:
It seems the current behavior is intentional, this PR breaks the tests.

boxed:
I think it might be a bit much to say it's intentional just because the tests actually check it :P The tests could have been black box tests and then they just set in stone what the current behavior is.

emontnemery:
Sure, but this PR does not adjust the tests.

mkenigs:
I fixed the unit test. There's still 2 broken but those appear to be broken on master. Explanation from commit message:

The current unit test is wrong because the UTC time is 2:00:00, so GMT5
should be 7:00:00, not 3:00:00

I'm not sure if there's more documentation, but I'm basing my understanding of `tz_offset` on the readme:
```
@freeze_time("2012-01-14 03:21:34", tz_offset=-4)
def test():
    assert datetime.datetime.utcnow() == datetime.datetime(2012, 1, 14, 3, 21, 34)
    assert datetime.datetime.now() == datetime.datetime(2012, 1, 13, 23, 21, 34)

    # datetime.date.today() uses local time
    assert datetime.date.today() == datetime.date(2012, 1, 13)
```

The first argument to `freeze_time` is the UTC time, and adding the offset gives the local time. So in the unit test the UTC time is 2:00:00

mkenigs:
@boxed @emontnemery does this look okay now that I fixed the test?

codejedi365:
@mkenigs, well done! this is the fix I needed as well because I was magically a double timezone offset away.  @spulec, please merge & release when you can.

PR #408, opened 2021-08-21
Remove several Python 2 workarounds

jdufresne:
Python 2 has not been supported since
2bea88ff771bc195029599d154c1bb9d0a02f2a2.


boxed:
Thanks!

PR #409, opened 2021-08-22
Remove use of unmaintained nose

jdufresne:
The nose project has ceased development. The last commit is from Mar 3,
2016. From their docs page:

https://nose.readthedocs.io/

> Note to Users
>
> Nose has been in maintenance mode for the past several years and will
> likely cease without a new person/team to take over maintainership.
> New projects should consider using Nose2, py.test, or just plain
> unittest/unittest2.

Simplify tests by using the stdlib unittest. One fewer dependency.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,9 +11,6 @@ test:
 travis:
 	pytest --cov
 
-tdaemon:
-	tdaemon -t nose ./tests/ --custom-args="--with-growl"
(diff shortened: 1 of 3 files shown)


jdufresne:
Fixed merge conflict.

boxed:
Thanks!

PR #419, opened 2021-08-26
Move type annotations inline and complete them

jdufresne:
Now that the project is Python 3.6+, it can safely use inline type
annotations. Doing so helps ensure that the types remain accurate and
useful both internally and both projects that consume freezegun. For
example, the pip project currently ignores missing freezegun types:

https://github.com/pypa/pip/blob/135faabfd62f8c3566fd4f99a5688316302cf1c8/tests/unit/test_self_check_outdated.py#L7

The CI system runs mypy on all support Python platforms as the stdlib
API can change between releases. The code has been altered to use
PEP-484 compliant version and platform checking as outlined at:
https://www.python.org/dev/peps/pep-0484/#version-and-platform-checking

The type checker already found a couple of mistakes with internal API
usage. This should help catch bugs in new pull requests earlier during
CI.

The py.typed file is now installed alongside the package to be compliant
with PEP-561:
https://www.python.org/dev/peps/pep-0561/#packaging-type-information

Using the spec "type: ignore" comment and cast calls were only done as a
last resort when the type checker couldn't be satisfied using more
conventional approaches.

By switching to using the setuptools explicit package_data approach, the
MANIFEST.in file was simplified.


bblommers:
This has been superceded by #543

Issue #398, opened 2021-05-07
Python 3.10: Deprecation warnings regarding event loop


tirkarthi:
Related to https://github.com/python/cpython/pull/23554

```
tests/test_asyncio.py::test_time_freeze_coroutine
  /root/checked_repos/freezegun/tests/test_asyncio.py:17: DeprecationWarning: There is no current event loop
    asyncio.get_event_loop().run_until_complete(frozen_coroutine())

tests/test_asyncio.py::test_time_freeze_async_def
  <string>:5: DeprecationWarning: There is no current event loop

-- Docs: https://docs.pytest.org/en/stable/warnings.html

```

PR #421, opened 2021-10-16
tests: asyncio: Fix Python 3.10 compatibility

jktjkt:
Since Python 3.10, the `asyncio.get_event_loop()` has been deprecated in favor of `asyncio.get_running_loop()`. That one, however, issues a warning when there's no event loop running (such as in this test suite).

Fix this by always requesting a new event loop. These methods have been available since at least Python 3.5 (and there are no older docs online at this point).

Bug: https://github.com/spulec/freezegun/issues/398

### DIFF
--- a/tests/test_asyncio.py
+++ b/tests/test_asyncio.py
@@ -14,7 +14,7 @@ def test_time_freeze_coroutine():
     async def frozen_coroutine():
         assert datetime.date.today() == datetime.date(1970, 1, 1)
 
-    asyncio.get_event_loop().run_until_complete(frozen_coroutine())
+    asyncio.new_event_loop().run_until_complete(frozen_coroutine())
 
 
 def test_time_freeze_async_def():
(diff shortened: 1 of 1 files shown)


tests/test_asyncio.py
```diff
@@ -14,7 +14,7 @@ def test_time_freeze_coroutine():
     async def frozen_coroutine():
         assert datetime.date.today() == datetime.date(1970, 1, 1)
 
-    asyncio.get_event_loop().run_until_complete(frozen_coroutine())
+    asyncio.new_event_loop().run_until_complete(frozen_coroutine())
```
graingert, line 17:
```suggestion
    asyncio.run(frozen_coroutine())
```

tests/test_asyncio.py
```diff
@@ -14,7 +14,7 @@ def test_time_freeze_coroutine():
...
 
 
 def test_time_freeze_async_def():
```
graingert, line 20: this test can just be removed because it's the same as the above test_time_freeze_coroutine

Issue #422, opened 2021-10-31
`datetime_to_fakedatetime` function ignores `fold` parameter


alexander-doroshenko:
since python 3.6 version to datetime class was added `fold` argument to disambiguate the times during a repeated interval.

An example of where it can hurt :
```python3
from datetime import timezone, datetime
from freezegun import api

d = datetime(2021, 10, 30, 23, tzinfo=timezone.utc)
isr = tz.gettz('Israel')
d.astimezone(isr).dst() == api.datetime_to_fakedatetime(d).astimezone(isr).dst()
>> False
```

couling:
In real terms this is particularly frustrating because we can't test TZ changes in our code. This is one of the big use cases for us.

```python
>>> from datetime import datetime
>>> from zoneinfo import ZoneInfo
>>>
>>> from freezegun import freeze_time
>>>
>>> UTC = ZoneInfo("UTC")
>>> UK = ZoneInfo("Europe/London")
>>>
>>> before_change = datetime(2023, 10, 29, 0, 30, tzinfo=UTC)
>>> with freeze_time(before_change):
...     print(before_change.astimezone(UK))
...     print(datetime.now(UTC).astimezone(UK))
...
2023-10-29 01:30:00+01:00
2023-10-29 01:30:00+01:00
>>> after_change = datetime(2023, 10, 29, 1, 30, tzinfo=UTC)
>>> with freeze_time(after_change):
...     print(after_change.astimezone(UK))
...     print(datetime.now(UTC).astimezone(UK))
...
2023-10-29 01:30:00+00:00
2023-10-29 01:30:00+01:00
```

Otto-AA:
I also ran into this issue. For me the conversion of `now.astimezone(gettz('CET')).astimezone(UTC) == now` failed because the `fold` parameter is not considered, so I cannot test this edge case (which is important in my application).

The PR can still be rebased from master without any conflicts and test errors and fixes my use case. Would it be possible to take a look, @bblommers ? I understand if you do not have time to do so, I just did not want to let this issue+PR get lost unnoticed, as it seems useful and quick to review.

PR #424, opened 2021-11-02
Add basic fold support

emontnemery:
Add basic fold support to enable testing of DST transitions with python 3.6+ when using a fold aware timezone library

Because fold information is lost when adding a `timedelta` to a `datetime`, this does not work when `tz_offset` is used.

Fixes https://github.com/spulec/freezegun/issues/422

### DIFF
--- a/freezegun/api.py
+++ b/freezegun/api.py
@@ -295,7 +295,8 @@ def datetime_to_fakedatetime(datetime):
                         datetime.minute,
                         datetime.second,
                         datetime.microsecond,
-                        datetime.tzinfo)
+                        datetime.tzinfo,
+                        fold=datetime.fold)
 
 
 def date_to_fakedate(date):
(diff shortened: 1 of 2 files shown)


couling:
Just ran into this bug, is there a reason it's not been merged, does it need further work?

freezegun/api.py
```diff
@@ -385,9 +386,14 @@ def timestamp(self):
...
         if tz:
-            result = tz.fromutc(now.replace(tzinfo=tz)) + cls._tz_offset()
+            result = tz.fromutc(now.replace(tzinfo=tz))
         else:
-            result = now + cls._tz_offset()
+            result = now
+
+        # Add the _tz_offset only if it's non-zero to preserve fold
+        if cls._tz_offset():
+            result += cls._tz_offset()
```
SoulofAkuma, line 395: You can add the timedelta and preserve the fold by first converting to UTC, then adding the timedelta and then converting back. Adding a timedelta offset while having a timezone argument as well makes relatively little sense though

SoulofAkuma:
@spulec Could this feature be merged? It would be very handy to test DST transitions

Issue #331, opened 2020-02-07
`ignore` doesn't seem to work in more complicated setup (Django + S3 / boto)


radoh:
I am using Django with S3 as the file storage (which uses boto3/botocore libs). I have a "management" command which creates some Django instances - I use freezegun, so these instances appear to be created in the past. However some of these models contain files, that are being saved to S3 - which raises some exceptions:
```
...
  File "/home/rado/.virtualenvs/twisto/lib/python3.6/site-packages/botocore/client.py", line 314, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/home/rado/.virtualenvs/twisto/lib/python3.6/site-packages/botocore/client.py", line 612, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (RequestTimeTooSkewed) when calling the ListObjectsV2 operation: The difference between the request time and the current time is too large.
```
this is due to some authentication code in `botocore.auth` module, that checks for time differences. I tried to solve this issue using `ignore` parameter (`freeze_time(datetime_in_past, ignore=['botocore'])`), but it didn't help, `botocore.auth` still used the `FakeDatetime` (I used debugger with breakpoints to pinpoint the issue).

I tried to recreate the issue with a simple reproducible example, so I came up with following:
```python
# test.py
from freezegun import freeze_time
from datetime import datetime
import stuff_in_between

import test_time #  <-- note this unused import

with freeze_time(datetime(2019, 1, 1, 1, 1, 1, 1), ignore=['test_time']):
    print(f'frozen time: {datetime.now()}')
    stuff_in_between.call_test_time()
```
```python
# stuff_in_between.py
def call_test_time():
    import test_time
    test_time.test_time()
```
```python
# test_time.py
from datetime import datetime

def test_time():
    print(f'real datetime.now: {datetime.now()}')
```
This code correctly works, freeze_time ignore `test_time` and it prints `real datetime.now: <real time>`. However if I omit the marked unused import in `test.py`, it doesn't work - freeze_time doesn't `ignore` the module and the script prints out `real datetime.now: <fake time>`.

I thought I could try out something similar in my real Django go, so I did following:
```python
...
import botocore
import botocore.auth
# hell, try to import everything possible related to S3
with freeze_time(datetime_in_past, ignore=['botocore']:
   ...
   # create some instances, that also save files to S3
```
[...]

PR #430, opened 2021-11-19
fix ignoring packages that are imported after freezing

### LINKED ISSUE #420: How to use Ignore packages
While I use freezegun with google storage api, I got the following error.

`google.auth.exceptions.RefreshError: ('invalid_grant: Invalid JWT: Token must be a short-lived token (60 minutes) and in a reasonable timeframe. Check your iat and exp values in the JWT claim.', {'error': 'invalid_grant', 'error_description': 'Invalid JWT: Token must be a short-lived token (60 minutes) and in a reasonable timeframe. Check your iat and exp values in the JWT claim.'})`

I think [ignore packages](https://github.com/spulec/freezegun#ignore-packages) may solve the problem. Ref: https://github.com/spulec/freezegun/pull/185

> Sometimes it's desired to ignore FreezeGun behaviour for particular packages (i.e. libraries). It's possible to ignore them for a single invocation:

I tried add ["google", "google.auth", "google.cloud"] in ignore list, but still got the same error.

```
from google.cloud import storage
import freezegun

ig_ms = ["google", "google.oauth2", "urllib3", "google.cloud"]
freezegun.configure(extend_ignore_list=ig_ms)


with freezegun.freeze_time("2017-05-21", ignore=ig_ms):
    client = storage.Client()

    print(list(client.list_buckets()))
```

I am confused about how to use Ignore Packages correctly.
For example: 

```
import urllib3
import freezegun

with freezegun.freeze_time("2017-05-21", ignore=['urllib3']):
    http = urllib3.PoolManager()
    resp = http.request("GET", "https://httpbin.org/robots.txt")
    print(resp.status)

```

whether I add urllib3 to ignore list or not, 
it will still raise `SystemTimeWarning: System time is way off (before 2020-07-01). This will probably lead to SSL verification errors`

pegler:
When a package is imported after freezegun starts, ignore would not work correctly. This fixes `FakeDate.today()`, `FakeDatetime.now()` and `FakeDatetime.utcnow()` to return the real date/datetime in ignored packages.

fixes #420


pegler:
hi @boxed and @spulec, thanks for your efforts in maintaining this package. We've found it incredibly useful, but unfortunately identified this bug. Right now we're patching freezegun to make it work correctly and it'd be great if we could get the fix applied to the package itself. Let me know if there are any changes you'd like to see. Thanks!

immerrr:
Hi,

Just ran into this because S3 was rejecting requests made from within `freeze_time` block as too old, and was unable to `ignore` my way around it. Turns out, `boto3` uses `import datetime; datetime.datetime` instead of `from datetime import datetime`, and that is enough for ignores to be _ignored_.

It is easy to show that this is broken with a small patch against existing tests:

https://github.com/spulec/freezegun/commit/5c305e6a4751f8224ea8763cd7ae283fca1123fd

And I came up with a similar fix:

https://github.com/spulec/freezegun/commit/7fa651590432212a5d63614a9abdddaf8f8e4a5d

Would be great to have this fixed upstream :+1:  Thanks!

immerrr:
@boxed would you be able to have a look at this? it is still impossible to use `freezegun` with any code that's uploading files to S3, because of #331, and this PR would fix it.

pegler:
@immerrr I switched to https://github.com/adamchainz/time-machine a while ago and have been very happy. You need to take a slightly different approach to handling ignores, but it worked well for our project.

freezegun/api.py
```diff
@@ -322,7 +322,11 @@ def __sub__(self, other):
 
     @classmethod
     def today(cls):
-        result = cls._date_to_freeze() + cls._tz_offset()
+        if _should_use_real_time():
+            result = real_date.today() + cls._tz_offset()
+        else:
+            result = cls._date_to_freeze() + cls._tz_offset()
```
boxed, line 328: I don't understand what this added logic is for. `cls._date_to_freeze()` calls `get_current_time()` which already calls `_should_use_real_time()` and does this type of thing so this change seems redundant?

https://github.com/spulec/freezegun/blob/92624e42130684163040b30a95a2c2e9e1cd7a4d/freezegun/api.py#L174

boxed:
@pegler that's cool! I've been meaning to try it, just haven't had time to check it out. 

As everyone here is painfully aware, neither me not spulec have the energy/motivation to maintain freezegun to the level it deserves as a very widely used library. So if time-machine is actually a viable replacement, that's very good to hear.

freezegun/api.py
immerrr: @boxed I'm happy to submit my change which [should be smaller](https://github.com/spulec/freezegun/compare/master...immerrr:consult-ignore-list-in-datetime-now-date-today?expand=1) if that would make it easier to get this in.

I could add the test case from this PR to make sure it covers the issue mentioned in this PR, too.

immerrr: Just to make sure we are talking about the same thing, you mean my patch that touches these two conditionals is too complex, right?

image

boxed: Yes. Correct. I feel like this logic should be fully contained within get_current_time itself. It just wouldn't be DRY otherwise, and it would also mean there's a ton of duplicate code and every place there isn't duplication would be a bug.

immerrr: There are also performance considerations here, e.g. some blocks like 
```
if _should_use_real_time():
    return call_some_builtin_function()
```
may work much faster than converting the result of `get_current_time()` back to the required format.

DRY is a valid principle, but repetition only manifests itself if the code needs to change. If it is a small module that is feature complete, what difference does it make if the code repeats one time, two times or ten times, the interpreter doesn't care.

That said, let me see if I can hide `should_use_real_time` inside `get_current_time`...

boxed: > may work much faster than converting the result of get_current_time() back to the required format.

There is no conversion in this code though.

immerrr: > There is no conversion in this code though.

There are conversions in functions like `fake_monotonic` whose "faking" versions use `get_current_time`

https://github.com/spulec/freezegun/blob/be4127bff37682591b0b099b9a2904689133969b/freezegun/api.py#L221-L225

> but get_current_time already calls should_use_real_time! 

I don't think it does, or maybe i'm looking somewhere else

https://github.com/spulec/freezegun/blob/be4127bff37682591b0b099b9a2904689133969b/freezegun/api.py#L169-L170

PR #441, opened 2022-02-14
Fix time cut-off on time_ns

pbabics:
Hello,

currently `time.time_ns` is being updated to time cut to seconds, which is kind of weird behavior in regards to `time.time` which returns the decimal part.

Basically this should (hopefully) fix the issue with:
`time.time_ns() / 1e9 != time.time()`


freezegun/api.py
```diff
@@ -180,7 +180,8 @@ def fake_time():
     def fake_time_ns():
         if _should_use_real_time():
             return real_time_ns()
-        return int(int(fake_time()) * 1e9)
+        current_time = get_current_time()
+        return int(calendar.timegm(current_time.timetuple()) * 1_000_000_000 + current_time.microsecond * 1_000)
```
marcinsulikowski, line 185: Why not `return int(fake_time() * 1e9)`?

marcinsulikowski: Oh, I see – `fake_time() * 1e9` could cause problems with precision.

Issue #439, opened 2022-01-19
as_kwarg is recognized as unexpected argument by PyCharm and SonarQube


fannyhub:
![freezegun](https://user-images.githubusercontent.com/5873729/150100521-93760414-68cc-4313-9b81-ce385e6cd954.PNG)

The function freeze_time works as expected but SQ and PyCharm mark one of its arguments (as_kwarg) as unexpected. Furthermore as_kwarg is not displayed by PyCharm on the list of the function's parameters.

PR #453, opened 2022-06-14
Fix as_kwarg type annotations

joey-rescale:
This fixes mypy errors and is probably also relevant to #439.

### DIFF
--- a/freezegun/api.pyi
+++ b/freezegun/api.pyi
@@ -30,7 +30,7 @@ class _freeze_time:
         ignore: Sequence[str],
         tick: bool,
         as_arg: bool,
-        as_kwarg: bool,
+        as_kwarg: str,
         auto_tick_seconds: float,
     ) -> None: ...
     @overload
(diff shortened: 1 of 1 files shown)


boxed:
Why str? It seems to me like it should be bool.

joey-rescale:
https://github.com/spulec/freezegun/blob/master/freezegun/api.py#L785
It's an index into a dict, hence str. (Unlike `as_arg` which is a bool)
Also see the default value on line 796 which is `''`.

boxed:
Ah, my bad. Thanks.

joey-rescale:
Thanks! There's no rush on my end (I'm using `type: ignore`), but for my own knowledge, what's the release cadence for this package?

boxed:
Totally random and very seldom.

joey-rescale:
Fair enough, thanks!

JOEY MARIANER

Senior Backend Engineer

Rescale ▪ High Performance Computing Built for the Cloud ▪ rescale.com
<http://www.rescale.com>

33 New Montgomery St. ▪ San Francisco, CA 94105

***@***.***

PR #454, opened 2022-06-22
Updating typing from types-freezegun 1.1.10

staticdev:
There are some old definitions for freezegun type-hints causing errors. This updates them.


staticdev:
@boxed could we add this also to next release?

staticdev:
@boxed can we get a release with this one?

boxed:
@staticdev done

Issue #382, opened 2021-01-19
TypeError: fake_time() takes 0 positional arguments but 1 was given


MuslimBeibytuly:
python version - 3.8.0
freezegun==1.0.0
when calling time.time() it throws TypeError: fake_time() takes 0 positional arguments but 1 was given

boxed:
You need to give at least the full stack trace. It would also help seeing your code that triggers this.

oaosman84:
Here's a stack trace from my end:
```
ERROR    django.request:log.py:224 Internal Server Error: /graphql
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
(TRUNCATED INTERNAL CODE)
  File "/usr/local/lib/python3.9/site-packages/rest_framework/throttling.py", line 124, in allow_request
    self.now = self.timer()
TypeError: fake_time() takes 0 positional arguments but 1 was given
```
Interestingly this happens only if I try to run one individual test. If I run the whole test suite (that includes the individual test), the individual test passes, but if I run it alone, it fails.

xemxav:
I report exactly the same problem and stacktrace as @oaosman84. 
No problem when the tests is run with pytest or with the whole test suite with Pycharm but it fails when run individualy.

Kub-AT:
Same here but appears randomly when running whole test suite with pytest
(pytest  + pytest-xdist + pytest-randomly)


```
TypeError: fake_time() takes 0 positional arguments but 1 was given
self = <unittest.case._Outcome object at 0x7f8cfe4dd610>
test_case = <apps.organization.tests.views.test_get_email_from_invite_key.TestGetEmailBasedOnOrganizationInviteKeyView testMethod=test_should_raise_error_key_is_expired>
isTest = True
    @contextlib.contextmanager
    def testPartExecutor(self, test_case, isTest=False):
        old_success = self.success
        self.success = True
        try:
>           yield
/usr/local/lib/python3.9/unittest/case.py:59: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <apps.organization.tests.views.test_get_email_from_invite_key.TestGetEmailBasedOnOrganizationInviteKeyView testMethod=test_should_raise_error_key_is_expired>
[...]

PR #459, opened 2022-07-19
Handle rest_framework.throttling.SimpleRateThrottle 

JulianMaurin:
# Handle rest_framework.throttling.SimpleRateThrottle 

According to the issue: https://github.com/spulec/freezegun/issues/382.

I can confirm that the issue comes from an incompatibility with the DRF throttling view (see: [DRF code](https://github.com/encode/django-rest-framework/blob/3.13.1/rest_framework/throttling.py#L63)).

Once the `SimpleRateThrottle.timer` is patched with `fake_time`, executing `self.timer()` (equivalent to `timer(self)`) is working with `time.time` but not with `fake_time`:

## How to reproduce

```python
>>> import time
>>> class A:
...     timer = time.time
...     def test(self):
...             print(self.timer())
... 
>>> A().test()
1658255082.1697752
>>>
>>> from freezegun.api import fake_time
>>> class B:
...     timer = fake_time
...     def test(self):
...             print(self.timer())
... 
>>> B().test()
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 4, in test
TypeError: fake_time() takes 0 positional arguments but 1 was given
``` 

## About the fix

It was not that simple. Testing an API using the [test client](https://docs.djangoproject.com/en/4.0/topics/testing/tools/#the-test-client) will load the module after the patch that is performed by Freezgun. To reproduce the context of late loading I introduce the  `tests.fake_module_lazy` module (In order to add a test demonstrating the error).

Then, to fix the error I changed the signature of the `freezegun.api.fake_time` function, to accept any args/kwars (and so allow the `self.timer` usage at the origin of the issue).

I feel the change is not risky at all but I can miss something.

### DIFF
--- a/freezegun/api.py
+++ b/freezegun/api.py
@@ -170,7 +170,7 @@ def get_current_time():
     return freeze_factories[-1]()
 
 
-def fake_time():
+def fake_time(*args, **kwargs):
     if _should_use_real_time():
         return real_time()
     current_time = get_current_time()
--- a/tests/test_class_import.py
+++ b/tests/test_class_import.py
@@ -134,12 +134,15 @@ def test_fake_strftime_function():
     assert fake_strftime_function() == '2012'
 
 
-@freeze_time("2022-01-01 12:00:00")
+@freeze_time("2022-01-01")
 def test_fake_time_function_as_class_attribute():
+    local_time = datetime.datetime(2022, 1, 1)
+    utc_time = local_time - datetime.timedelta(seconds=time.timezone)
+    expected_timestamp = time.mktime(utc_time.timetuple())
+
     fake_module_lazy.load()
 
-    with pytest.raises(TypeError, match='fake_time\(\) takes 0 positional arguments but 1 was given'):
-        assert fake_module_lazy.TimeAsClassAttribute().call_time()
+    assert fake_module_lazy.TimeAsClassAttribute().call_time() == expected_timestamp
 
 
 def test_import_after_start():
--- a/tests/test_class_import.py
+++ b/tests/test_class_import.py
@@ -1,5 +1,8 @@
 import time
 import sys
+
+import pytest
+
 from .fake_module import (
     fake_date_function,
     fake_datetime_function,
@@ -9,6 +12,7 @@ from .fake_module import (
     fake_time_function,
 )
 from . import fake_module
+from . import fake_module_lazy
 from freezegun import freeze_time
 from freezegun.api import (
     FakeDatetime,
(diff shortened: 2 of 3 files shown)


tests/test_class_import.py
```diff
@@ -134,12 +134,15 @@ def test_fake_strftime_function():
     assert fake_strftime_function() == '2012'
 
 
-@freeze_time("2022-01-01 12:00:00")
+@freeze_time("2022-01-01")
 def test_fake_time_function_as_class_attribute():
+    local_time = datetime.datetime(2022, 1, 1)
+    utc_time = local_time - datetime.timedelta(seconds=time.timezone)
+    expected_timestamp = time.mktime(utc_time.timetuple())
```
JulianMaurin, line 141: This part is duplicated from `test_import_time`.
I can extract it to `_date_to_time(year: int, mont: int, day: int)` at the end of the module, wdyt ?

JulianMaurin: :heavy_check_mark: at 9d2b92f7faedec016ab17b088ebf330024ddf15f

JulianMaurin:
@staticdev, @boxed , 

Any opinion about this PR?

freezegun/api.py
```diff
@@ -170,7 +170,7 @@ def get_current_time():
     return freeze_factories[-1]()
 
 
-def fake_time():
+def fake_time(*args, **kwargs):
```
boxed, line 173: Why *args, **kwargs here? That seems very wrong to me.

JulianMaurin: Hello @boxed , 
Thanks for your review. May I humbly ask you to elaborate? 

I think I made an effort to explain the reproduction of the bug, both in the description and in the test.
If I'm off track, can you possibly guide me to a solution that you think is more appropriate?

staticdev: @JulianMaurin this is breaking the API typing on `api.pyi`, so users that have mypy will have problems. I also think you can add typed arguments to satisfy your needs with default `arguments=None` so it will also not break current applications and maintain the types well defined.

tests/test_class_import.py
```diff
@@ -130,6 +131,14 @@ def test_fake_strftime_function():
     assert fake_strftime_function() == '2012'
 
 
+@freeze_time("2022-01-01")
+def test_fake_time_function_as_class_attribute():
```
staticdev, line 135: It is great that you added tests, but it is also important to explain on `README` the new optional usage in English so more people can easily reuse your solution =)

staticdev: You can also add some description of this change on `CHANGELOG`.

staticdev (changes requested):

freezegun/api.py
JulianMaurin: I thought that this version does not break the application, and in addition  `fake_time` function is not defined in `api.pyi`, but maybe I did not understand something.

Would you prefer this version (with a docstring ofc):
```suggestion
def fake_time(_ : Any):
```

JulianMaurin (commented):
Hey @staticdev ,
Thank you for taking the time to develop. 

I will add an entry to the changelog if we continue this PR, thanks for the suggestion :bow:

However I don't see what should be added to the README, can you be more precise?

It would be nice to fix this bug, but if you think this is not the right way at all, tell me directly and I will close the PR and we will all save time.

PR #460, opened 2022-08-01
Don't freeze pytest timings

### LINKED ISSUE #433: When freezegun is used at a class level, the pytest durations report shows invalid times
Here's a real example with fake paths illustrating the problem:
```
============================ slowest test durations ============================
1355132.16s teardown application/path/test_security_record
61.27s setup    application/path/test_viewing_client_statements
60.38s setup    application/path/test_start_date
60.23s setup    application/path/test_send
...
```

`1355132.16s` corresponds to over 15 days.

I suspect that the pytest durations report logic is relying on `datetime.datetime.now()` which is patched by `freezegun` resulting in the report showing incorrect durations.

I don't think this issue is w/in freezegun but it does result in the above reported undesirable side effect.

DavidS-cloud:
This avoids class-level decorator usage messing with pytest timings.

This fixes #433.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -30,7 +30,7 @@ Decorator
     def test():
         assert datetime.datetime.now() == datetime.datetime(2012, 1, 14)
 
-    # Or a unittest TestCase - freezes for every test, from the start of setUpClass to the end of tearDownClass
+    # Or a unittest TestCase - freezes for every test, and set up and tear down code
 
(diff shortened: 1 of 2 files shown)


DavidS-cloud:
I've now deployed this change into our CI, and it has passed and is now keeping more sensible time.

boxed:
Nice

PR #369, opened 2020-09-30
Add support for `time.monotonic` (and `time.monotonic_ns`)

jmehnle:
There's a seemingly abandoned attempt at adding support for `time.monotonic` in https://github.com/spulec/freezegun/pull/199. As there were major changes to `freezegun` since then, I decided to take a fresh stab at implementing such support. It's worth noting that `time.monotonic` doesn't give many guarantees to begin with, so not many are tested.

I'd be happy to add a few documentation bits if this implementation and the accompanying tests work for you.

PR #470, opened 2022-09-14
Improve asyncio support to avoid hangs of asyncio.sleep()

### LINKED ISSUE #401: freezegun hangs parametrized asyncio test
Hi,

First of all, thank you for creating this project. It's a real help to test time-related code in an easy an consistent way. Thank you.

I'm currently trying to debug why some of my tests hang forever. It looks like it's a combination of `freezegun`, parametrized tests and `asyncio`. Minimal test case:

```python
import asyncio
import pytest
import freezegun


@pytest.fixture(params=[1, 2])
def number(request):
    """Returns a number"""
    return request.param


async def do_stuff(num):
    await asyncio.sleep(0.001)
    return num


@pytest.mark.asyncio
async def test_do_stuff(number):
    with freezegun.freeze_time() as frozen_datetime:
        coro = do_stuff(number)
        result = await coro
        assert result == number
```

The first run (with `number == 1`) runs fine, but the second run (with `number == 2`) hangs at the `await coro` step:

```
$ pytest -v tests/freezegun_test.py
================================ test session starts =================================
platform darwin -- Python 3.9.1, pytest-6.2.4, py-1.10.0, pluggy-0.13.1 -- /<redacted>/venv/bin/python3
cachedir: .pytest_cache
rootdir: /<redacted>
plugins: asyncio-0.15.1
collected 2 items                                                                    

tests/freezegun_test.py::test_do_stuff[1] ^C

!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! KeyboardInterrupt !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
/usr/local/Cellar/python@3.9/3.9.1_3/Frameworks/Python.framework/Versions/3.9/lib/python3.9/selectors.py:562: KeyboardInterrupt
(to show a full traceback on KeyboardInterrupt use --full-trace)
=============================== no tests ran in 0.03s ================================
Task was destroyed but it is pending!
task: <Task pending name='Task-1' coro=<test_do_stuff() done, defined at /<redacted>/tests/freezegun_test.py:17> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x102539a90>()]>>
```

Other combinations work:
* Running without freezegun
* Running a single test (non-parametrized, or parametrized with a single run)
* non-asyncio calls

How can I help to identify and fix this bug?

<details>
<summary>Version details</summary>
$ python -V
Python 3.9.1

$ pip list
Package         Version
--------------- -------
attrs           21.2.0
freezegun       1.1.0
iniconfig       1.1.1
packaging       20.9
pip             20.2.3
pluggy          0.13.1
py              1.10.0
pyparsing       2.4.7
pytest          6.2.4
pytest-asyncio  0.15.1
python-dateutil 2.8.1
setuptools      49.2.1
six             1.16.0
toml            0.10.2
WARNING: You are using pip version 20.2.3; however, version 21.1.2 is available.
You should consider upgrading via the '/Users/niels/nextcloud/src/velbuspy/venv/bin/python3 -m pip install --upgrade pip' command.
</details>

### LINKED ISSUE #437: calling (and awaiting) asyncio.sleep seems to cause an infinite hang on 1.1.0
I upgraded a project to the latest version of freezegun recently and a test which relied on an interaction which happened in the background started failing.  I tracked it down to being due to any call to `await asyncio.sleep(..)` to hang forever.  Perhaps the internal mechanism used in asyncio is sensitive to the way freezegun patches time modules?

I think this may be related to: https://github.com/spulec/freezegun/issues/290 but I am not sure since this works fine on version 1.0.0 which was released after that issue was created.  Additionally I wanted to submit a simple example to reproduce the problem.  If it's indeed the same issue, my apologies, I am happy to contribute over there instead.

My python version is: 3.9.1

Here's a proof of concept (t.py):

```python
import signal
import sys
import asyncio

import freezegun


TIMEOUT = 5


def timeout(sig, frame):
    print(f"exiting due to {TIMEOUT}s timeout")
    sys.exit(1)


async def main():
    print("Normal asyncio.sleep(1)")
    await asyncio.sleep(1)
    print("Done!")
    with freezegun.freeze_time("2022-01-01"):
        print("Frozen asyncio.sleep(1)")
        await asyncio.sleep(1)
    print("Done! (success)")


if __name__ == "__main__":
    signal.signal(signal.SIGALRM, timeout)
    signal.alarm(TIMEOUT)
    asyncio.run(main())
```

Running this on 1.1.0 causes the signal handler to execute, since the main method never gets passed the `asyncio.sleep(1)` call within the freeze_time context.

```
$ venv/bin/pip install freezegun==1.1.0 && venv/bin/python t.py
Requirement already satisfied: freezegun==1.1.0 in ./venv/lib/python3.9/site-packages (1.1.0)
Requirement already satisfied: python-dateutil>=2.7 in ./venv/lib/python3.9/site-packages (from freezegun==1.1.0) (2.8.2)
Requirement already satisfied: six>=1.5 in ./venv/lib/python3.9/site-packages (from python-dateutil>=2.7->freezegun==1.1.0) (1.16.0)
WARNING: You are using pip version 20.3.3; however, version 21.3.1 is available.
You should consider upgrading via the '/Users/ses/w/freezegun-async-sleep-poc/venv/bin/python -m pip install --upgrade pip' command.
Normal asyncio.sleep(1)
Done!
Frozen asyncio.sleep(1)
exiting due to 5s timeout
```

Downgrading to 1.0.0 seems to "fix" this.

```
$ venv/bin/pip install freezegun==1.0.0 && venv/bin/python t.py
Collecting freezegun==1.0.0
  Using cached freezegun-1.0.0-py2.py3-none-any.whl (14 kB)
Requirement already satisfied: python-dateutil>=2.7 in ./venv/lib/python3.9/site-packages (from freezegun==1.0.0) (2.8.2)
Requirement already satisfied: six>=1.5 in ./venv/lib/python3.9/site-packages (from python-dateutil>=2.7->freezegun==1.0.0) (1.16.0)
Installing collected packages: freezegun
  Attempting uninstall: freezegun
    Found existing installation: freezegun 1.1.0
    Uninstalling freezegun-1.1.0:
      Successfully uninstalled freezegun-1.1.0
Successfully installed freezegun-1.0.0
WARNING: You are using pip version 20.3.3; however, version 21.3.1 is available.
You should consider upgrading via the '/Users/ses/w/freezegun-async-sleep-poc/venv/bin/python -m pip install --upgrade pip' command.
Normal asyncio.sleep(1)
Done!
Frozen asyncio.sleep(1)
Done! (success)
```

marcinsulikowski:
### Clean up asyncio tests

We finish the cleanup of `test_asyncio.py` done in c60dd14b05f09c482d4
and in 204647b305 by removing remnants of support of Pythons with no
`asyncio` module because all Python interpreters supported by freezegun
already support `asyncio`.

### Avoid `asyncio.sleep()` hanging forever when time is frozen

The following code:

    async def test():
        with freeze_time("2020-01-01"):
            await asyncio.sleep(0.01)

hangs forever since FreezeGun 1.1.0 because 1.1.0 started patching
`time.monotonic()` (see #369) which is used internally by `asyncio`
event loops to schedule code for execution in the future. This breaks
many projects that uses FreezeGun to test asynchronous code.

We fix this by changing `freeze_time` to patch asyncio event loop's
`time()` method in a way that it uses real monotonic time instead of the
frozen one. Note that we couldn't achieve this by adding `asyncio` to
`DEFAULT_IGNORE_LIST` in `freezegun/config.py` because any running async
code has functions from the `asyncio` module on its stack -- adding
`asyncio` to the ignore list would just disable freezing time in any
async code. This is why we patch one method of a specific class instead.

This change not only fixes `asyncio.sleep()` but also things like
`asyncio.get_running_loop().call_later` (for scheduling task execution
in the future) which in turn makes things like timeouts work in async
code while time is frozen. This may not be desired because some users
may expect that execution of events scheduled to happen in the future
can be controlled using FreezeGun. However, it's not easy to distinguish
between things that users would like to see frozen time and those which
should not (like `asyncio.sleep()`) because all of them use the same
clock. Therefore, we opt for making all `asyncio` internals not affected
by FreezeGun.

We also add more tests that verify how FreezeGun interacts with asyncio
code, including tests that cover the scenario described in #437 which we
aim to fix.

Closes #401
Closes #437

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -20,3 +20,4 @@ Patches and Suggestions
 - `Lukasz Balcerzak <https://github.com/lukaszb>`_
 - `Hannes Ljungberg <hannes@5monkeys.se>`_
 - `staticdev <staticdev-support@proton.me>`_
+- `Marcin Sulikowski <https://github.com/marcinsulikowski>`_
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,11 @@
 Freezegun Changelog
 ===================
 
+1.3.0
+-----
+
+* Fixed `asyncio` support to avoid `await asyncio.sleep(1)` hanging forever.
+
 1.2.2
 -----
 
--- a/freezegun/api.py
+++ b/freezegun/api.py
@@ -1,5 +1,6 @@
 from . import config
 from ._async import wrap_coroutine
+import asyncio
 import copyreg
 import dateutil
 import datetime
@@ -726,6 +727,21 @@ class _freeze_time:
                         setattr(module, attribute_name, fake)
                         add_change((module, attribute_name, attribute_value))
 
+        # To avoid breaking `asyncio.sleep()`, let asyncio event loops see real
+        # monotonic time even though we've just frozen `time.monotonic()` which
+        # is normally used there. If we didn't do this, `await asyncio.sleep()`
+        # would be hanging forever breaking many tests that use `freeze_time`.
+        #
+        # Note that we cannot statically tell the class of asyncio event loops
+        # because it is not officially documented and can actually be changed
+        # at run time using `asyncio.set_event_loop_policy`. That's why we check
+        # the type by creating a loop here and destroying it immediately.
+        event_loop = asyncio.new_event_loop()
+        event_loop.close()
+        EventLoopClass = type(event_loop)
+        add_change((EventLoopClass, "time", EventLoopClass.time))
+        EventLoopClass.time = lambda self: real_monotonic()
+
         return freeze_factory
 
     def stop(self):
@@ -739,8 +755,8 @@ class _freeze_time:
             datetime.date = real_date
             copyreg.dispatch_table.pop(real_datetime)
             copyreg.dispatch_table.pop(real_date)
-            for module, module_attribute, original_value in self.undo_changes:
-                setattr(module, module_attribute, original_value)
+            for module_or_object, attribute, original_value in self.undo_changes:
+                setattr(module_or_object, attribute, original_value)
             self.undo_changes = []
 
             # Restore modules loaded after start()
--- a/tests/test_asyncio.py
+++ b/tests/test_asyncio.py
@@ -1,5 +1,6 @@
 import asyncio
 import datetime
+import time
 
 from freezegun import freeze_time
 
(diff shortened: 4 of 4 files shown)


marcinsulikowski:
I rebased this pull request on top of the current `master` branch and resolved all the conflicts. @boxed, what do you think about it?

Issue #344, opened 2020-04-01
Using tz_offset breaks t == fromtimestamp(t).timestamp()


Strilanc:
```python
import freezegun

def test_without_freezegun():
    t = datetime.datetime.now().timestamp()
    t2 = datetime.datetime.fromtimestamp(t).timestamp()
    assert t == t2  # PASS


@freezegun.freeze_time(datetime.datetime.fromtimestamp(100_000), tz_offset=-1)
def test_with_freezegun():
    t = datetime.datetime.now().timestamp()
    t2 = datetime.datetime.fromtimestamp(t).timestamp()
    assert t == t2  # FAIL
```

PR #475, opened 2022-10-27
make sure fromtimestamp always return FakeDateTime

Koswu:
This fix https://github.com/spulec/freezegun/issues/344

This is my first PR, if something goes wrong, please let me know, PTAL

### DIFF
--- a/freezegun/api.py
+++ b/freezegun/api.py
@@ -374,10 +374,11 @@ class FakeDatetime(real_datetime, FakeDate, metaclass=FakeDatetimeMeta):
     @classmethod
     def fromtimestamp(cls, t, tz=None):
         if tz is None:
-            return real_datetime.fromtimestamp(
-                    t, tz=dateutil.tz.tzoffset("freezegun", cls._tz_offset())
-                ).replace(tzinfo=None)
-        return datetime_to_fakedatetime(real_datetime.fromtimestamp(t, tz))
+            tz = dateutil.tz.tzoffset("freezegun", cls._tz_offset())
+            result = real_datetime.fromtimestamp(t, tz=tz).replace(tzinfo=None)
+        else:
+            result = real_datetime.fromtimestamp(t, tz)
+        return datetime_to_fakedatetime(result)
 
     def timestamp(self):
         if self.tzinfo is None:
--- a/tests/test_datetimes.py
+++ b/tests/test_datetimes.py
@@ -93,6 +93,15 @@ def test_tz_offset():
     freezer.stop()
 
 
+def test_timestamp_tz_offset():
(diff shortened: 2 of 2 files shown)


boxed:
The proposed code change seems to be no change of behavior. Could you point out what it changes?

Koswu:
In the `fromtimestamp` method, when the `tz` parameter is None, It will return the real datetime object. This object's `.timestamp()` behavior will [depend on the system timezone](https://docs.python.org/3/library/datetime.html#datetime.datetime.timestamp), even it tzinfo is None.  For example: 
```
# my system timezone is UTC+8
>>> dt = datetime.datetime(1970, 1, 1, 8, 0, 1)
>>> dt.tzinfo
None
>>> dt.timestamp()
1.0
```

However, the fake one depend on the freeze time zone: 
```
>>> freezer = freeze_time(datetime.datetime(1970, 1, 1, 0, 0, 1), tz_offset=-1)
>>> freezer.start()
>>> datetime.datetime(1970, 1, 1, 8, 0, 1).timestamp()
32401.0
``` 

So if we make these steps:
1. make a fake datetime object
2. convert it to timestamp
3. convert it to datetime
4. convert it to timestamp

we will have a type chain like this: 
>  FakeDatetime -> int -> RealDatetime -> int

Since we know there are different behavior between `FakeDatetime.timestamp()` and `RealDatetime.timestamp()`, it will cause problem that two timestamp in the type chain are not equal. 

That's why I added the test.

freezegun/api.py
```diff
@@ -371,10 +371,11 @@ def astimezone(self, tz=None):
...
     def fromtimestamp(cls, t, tz=None):
         if tz is None:
-            return real_datetime.fromtimestamp(
-                    t, tz=dateutil.tz.tzoffset("freezegun", cls._tz_offset())
-                ).replace(tzinfo=None)
-        return datetime_to_fakedatetime(real_datetime.fromtimestamp(t, tz))
+            tz = dateutil.tz.tzoffset("freezegun", cls._tz_offset())
+            result = real_datetime.fromtimestamp(t, tz=tz).replace(tzinfo=None)
+        else:
+            result = datetime_to_fakedatetime(real_datetime.fromtimestamp(t, tz))
```
bblommers, line 377:
```suggestion
            result = real_datetime.fromtimestamp(t, tz)
```

The `datetime_to_fakedatetime` already happens at the end - no need to duplicate this call.

Happy to merge once that's fixed.

bblommers (changes requested):

freezegun/api.py
Koswu: fixed. thanks for pointing this out.

bblommers (approved):
Great - thanks for the fix @Koswu!

Issue #129, opened 2016-01-21
Freezegun and threading


sirex:
When freezegun is used together with threading, quite unexpected and hard to find things can happen.

Here is example:

```python
from Queue import Queue
from freezegun import freeze_time

@freeze_time('2016-01-04 16:49:59')
def test_me():
    queue = Queue()
    queue.get(block=True, timeout=0.1)
```

This will deadlock forever, even if `timeout=0.1` is specified, because `freezegun` patches `time`. Same thing happens for many other threading routines.

It is quite easy to fix:

```python
@freeze_time('2016-01-04 16:49:59', ignore=['threading', 'Queue'])
def test_me():
    queue = Queue()
    queue.get(block=True, timeout=0.1)
```

But at least for me it took quite a lot of time to figure this out.

I would suggest to add `threading` and `Queue` to `ignore` by default, or at least explicitly specify this in documentation.

spulec:
Thanks!

thomasnal:
I have just spent three hours tracing this issue after it triggered unexpectedly, likely due a version upgrade.   Very unexpected effect of freezegun and in my context the reasons for Queue timeout not working suddenly was not pointing to freezegun so was not able to find any help searching internet.  After damage has been done, hairs lost and reason found, tracing python libraries, I have come to ask to have multiprocessing ignored by default.

boxed:
@thomasnal that sounds reasonable to me. Could you make a PR for that? Should be a very simple change.

PR #489, opened 2023-02-02
Don't freeze multiprocessing

isra17:
Fix https://github.com/spulec/freezegun/issues/129#issuecomment-898942418


boxed:
Thanks. I do however recommend moving over to time-machine.

PR #403, opened 2021-06-21
Typeshed types

staticdev:
Closes #318
Closes #319
Closes #365

PR #509, opened 2023-08-12
better represent the return value of freeze_time at typing time

asottile:
currently to use any of the freezegun methods with a `freeze_time` context you need to assert `isinstance` after getting the return value

this change makes it so the yield types are correctly inferred:

```python
import freezegun

with freezegun.freeze_time('2023-01-01 00:00:00') as t1:
    print(t1)  # <freezegun.api.FrozenDateTimeFactory object at 0x7ffbe315e4a0>
    reveal_type(t1)  # note: Revealed type is "freezegun.api.FrozenDateTimeFactory"

with freezegun.freeze_time('2023-01-01 00:00:00', tick=True) as t2:
    print(t2)  # <freezegun.api.TickingDateTimeFactory object at 0x7ffbe308c6a0>
    reveal_type(t2)  # note: Revealed type is "freezegun.api.TickingDateTimeFactory"

with freezegun.freeze_time('2023-01-01 00:00:00', auto_tick_seconds=3) as t3:
    print(t3)  # <freezegun.api.StepTickTimeFactory object at 0x7ffbe25fa020>
    reveal_type(t3)  # note: Revealed type is "freezegun.api.StepTickTimeFactory"
```

### DIFF
--- a/freezegun/api.pyi
+++ b/freezegun/api.pyi
@@ -1,11 +1,12 @@
 from collections.abc import Awaitable, Callable, Iterator, Sequence
 from datetime import date, datetime, timedelta
 from numbers import Real
-from typing import Any, TypeVar, overload
+from typing import Generic, Literal, TypeVar, overload
 from typing_extensions import TypeAlias
 
 _T = TypeVar("_T")
 _Freezable: TypeAlias = str | datetime | date | timedelta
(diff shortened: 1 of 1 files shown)


asottile:
perhaps @staticdev / @boxed could take a look at this one since they reviewed / implemented #403

boxed:
I've moved to time-machine and I recommend everyone else to do the same.

asottile:
tyty -- we ended up abandoning freezegun too so I no longer need this

PR #511, opened 2023-08-24
Fix deprecation in Python 3.12

### LINKED ISSUE #508: utcnow is deprecated
With python 3.12 `datetime.datetime.utcnow` is deprecated. Used in _parse_time_to_freeze

Dreamsorcerer:
Fixes #508.

### DIFF
--- a/freezegun/api.py
+++ b/freezegun/api.py
@@ -470,7 +470,7 @@ def _parse_time_to_freeze(time_to_freeze_str):
     elif isinstance(time_to_freeze_str, datetime.date):
         time_to_freeze = datetime.datetime.combine(time_to_freeze_str, datetime.time())
     elif isinstance(time_to_freeze_str, datetime.timedelta):
-        time_to_freeze = datetime.datetime.utcnow() + time_to_freeze_str
+        time_to_freeze = datetime.datetime.now(datetime.UTC) + time_to_freeze_str
     else:
         time_to_freeze = parser.parse(time_to_freeze_str)
 


freezegun/api.py
```diff
@@ -470,7 +470,7 @@ def _parse_time_to_freeze(time_to_freeze_str):
     elif isinstance(time_to_freeze_str, datetime.date):
         time_to_freeze = datetime.datetime.combine(time_to_freeze_str, datetime.time())
     elif isinstance(time_to_freeze_str, datetime.timedelta):
-        time_to_freeze = datetime.datetime.utcnow() + time_to_freeze_str
+        time_to_freeze = datetime.datetime.now(datetime.timezone.utc) + time_to_freeze_str
```
cdce8p, line 473: I haven't checked it but this might have unintended consequences. A full replacement for `utcnow` would be `now(timezone.utc).replace(tzinfo=None)`. Otherwise you're converting a datetime object without a timezone to a tz aware object.

_It might work, but it could also fail. Just something to be aware of and test._

There is also at least one more `utcnow()` call just a few lines above in the same function
https://github.com/spulec/freezegun/blob/4f4496380deefceead7bef23bccaca17c2bdecfa/freezegun/api.py#L461-L466

Dreamsorcerer: Ah, I just went by the docs which say:
> As such, the recommended way to create an object representing the current time in UTC is by calling datetime.now(timezone.utc)

> Deprecated since version 3.12: Use [datetime.now()](https://docs.python.org/3.13/library/datetime.html#datetime.datetime.now) with [UTC](https://docs.python.org/3.13/library/datetime.html#datetime.UTC) instead.

Those docs should probably be clarified if it's not a safe replacement.

Dreamsorcerer: Added the missing ones, I'll take a closer look later at whether it needs be changed.

Dreamsorcerer: OK, I think this is fine. These 2 are converted to naive datetimes at the bottom of the function before it returns, so the result will end up the same anyway. The one in the class further up looks like an aware object should work just fine.

kurtmckee (approved):

kurtmckee:
@spulec I recommend merging this PR. It will resolve an outstanding DeprecationWarning in Python 3.12, which was released on Monday.

sirosen:
Hey, second person looking at the 3.12 breakage (full disclosure, @kurtmckee and I work together :slightly_smiling_face: )! I think the person whose attention we need is @boxed, who appears to have commit access on this repo.

boxed:
@sirosen I do not. I have switched to time-machine long ago and I recommend you do too.

sirosen:
I'm surprised/confused by this. The most recently merged PRs appear to be done by you in February, which is why I assumed you'd be able to help patch this.
If you don't have write access on the repo then it bodes ill for `freezegun`, but...

> I have switched to time-machine long ago and I recommend you do too.

... thanks for this recommendation! Looking into it for all of two minutes shows that it's from a maintainer with a decent track record for keeping projects healthy long-term, and it looks like nearly a drop-in replacement. I'll start looking into this for our team's needs.

boxed:
@sirosen This project has been functionally abandoned for several years. I tried to fix that for a while. Then I found out that time-machine exists.

Dreamsorcerer:
> I have switched to time-machine long ago and I recommend you do too.

We just tried this and it doesn't support pypy, which is a bit of a blocker to migrating over.

Dreamsorcerer:
> > I have switched to time-machine long ago and I recommend you do too.
> 
> We just tried this and it doesn't support pypy, which is a bit of a blocker to migrating over.

We decided to switch anyway, skipping the test on Pypy, but now we have reports from downstream packagers that time-machine segfaults when they run it. So far it is not looking like the best replacement for freezegun, and is a shame that this is not being maintained anymore...

kurtmckee:
Thanks @bblommers! I see you're listed as a maintainer on PyPI -- would you be willing to push a release to address this?

PR #518, opened 2023-12-04
Switch to pytest-cov

mtelka:
The `pytest-coverage` projects [suggests](https://pypi.org/project/pytest-coverage/0.0/): Use [pytest-cover](https://pypi.python.org/pypi/pytest-cover/) instead.

The [pytest-cover](https://pypi.org/project/pytest-cover/) says: pytest-cover has been merged back into [pytest-cov 2.0](https://pypi.python.org/pypi/pytest-cov).

So let's use `pytest-cov`.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 pytest
-pytest-coverage
+pytest-cov
 coveralls
 python-dateutil >= 2.7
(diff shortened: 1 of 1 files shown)


bblommers (approved):
Makes sense @mtelka - thanks for the PR!

PR #525, opened 2023-12-07
Only mock asyncio on request

### LINKED ISSUE #521: [Discussion] about freezing asyncio
PR https://github.com/spulec/freezegun/pull/470 basically unfroze asyncio during freeze_time.

Although i understand the need for it (not breaking tests that perform real waits on real events), it breaks the main purpose of freeze_time: Asyncio is no-longer frozen even without tick and i can no longer trigger events that would happen in an asyncio future without waiting the full duration( for instance, a common pattern is to do asyncio.sleep(remaining_time)). This makes some tests that were instant have to wait for whole timeout instead with freezegun>=1.3.0.

I understand that both properties may be desirable, but, the first one should only work with tick=True (it is currently inconditional), and may still manipulate monotonic for asyncio (it is desirable to trigger futures events with freezegun). 

A switch to get one behaviour or another may be needed; i guess there are even case where monotonic and time may need to be handled in a completely othogonal manner (as manipulating monotonic *will* always end up breaking a fundamental promise of the api: being monotonic).

So should i make a PR to make this behavior opt-in ? Add a new switch for asyncio ?

(Note, i have a longer take on this on a similar (but note quite the same) issue in time-machine: https://github.com/adamchainz/time-machine/pull/406)

### LINKED ISSUE #522: ``PermissionError: [Errno 1] Operation not permitted`` appears when upgrading to ``1.3.0``
I've not been able to pinpoint exactly what the issue is in freezegun (maybe async code, just a wild guess ?), but I bisected my CI failing to the upgrade from 1.2.2 to 1.3.0. Here's the stacktrace:
```
self = <_UnixSelectorEventLoop running=False closed=False debug=False>
fd = <MagicMock name='socket().fileno()' id='139676523736240'>
callback = <bound method BaseSelectorEventLoop._read_from_self of <_UnixSelectorEventLoop running=False closed=False debug=False>>
args = (), handle = <Handle BaseSelectorEventLoop._read_from_self()>
    def _add_reader(self, fd, callback, *args):
        self._check_closed()
        handle = events.Handle(callback, args, self, None)
        try:
>           key = self._selector.get_key(fd)
/usr/local/lib/python3.10/asyncio/selector_events.py:256: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <selectors.EpollSelector object at 0x7f08f9947d30>
fileobj = <MagicMock name='socket().fileno()' id='139676523736240'>
    def get_key(self, fileobj):
        """Return the key associated to a registered file object.
    
        Returns:
        SelectorKey for this file object
        """
        mapping = self.get_map()
        if mapping is None:
            raise RuntimeError('Selector is closed')
        try:
            return mapping[fileobj]
        except KeyError:
>           raise KeyError("{!r} is not registered".format(fileobj)) from None
E           KeyError: "<MagicMock name='socket().fileno()' id='139676523736240'> is not registered"
/usr/local/lib/python3.10/selectors.py:193: KeyError
During handling of the above exception, another exception occurred:
args = ()
kwargs = {'caplog': <_pytest.logging.LogCaptureFixture object at 0x7f08f9a983a0>, 'fake_fherd_input': <function fake_fherd_inpu..., 'path_to_directory': PosixPath('/builds/ukyu/zefzefz-fherd-bot/tests/cli/command/use_cases/ikd'), ...}
    def wrapper(*args, **kwargs):
>       with self as time_factory:
/usr/local/lib/python3.10/site-packages/freezegun/api.py:816: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/local/lib/python3.10/site-packages/freezegun/api.py:634: in __enter__
    return self.start()
/usr/local/lib/python3.10/site-packages/freezegun/api.py:739: in start
    event_loop = asyncio.new_event_loop()
/usr/local/lib/python3.10/asyncio/events.py:783: in new_event_loop
    return get_event_loop_policy().new_event_loop()
/usr/local/lib/python3.10/asyncio/events.py:673: in new_event_loop
    return self._loop_factory()
/usr/local/lib/python3.10/asyncio/unix_events.py:64: in __init__
    super().__init__(selector)
/usr/local/lib/python3.10/asyncio/selector_events.py:56: in __init__
    self._make_self_pipe()
/usr/local/lib/python3.10/asyncio/selector_events.py:107: in _make_self_pipe
    self._add_reader(self._ssock.fileno(), self._read_from_self)
/usr/local/lib/python3.10/asyncio/selector_events.py:258: in _add_reader
    self._selector.register(fd, selectors.EVENT_READ,
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <selectors.EpollSelector object at 0x7f08f9947d30>
fileobj = <MagicMock name='socket().fileno()' id='139676523736240'>, events = 1
data = (<Handle BaseSelectorEventLoop._read_from_self()>, None)
    def register(self, fileobj, events, data=None):
        key = super().register(fileobj, events, data)
        poller_events = 0
        if events & EVENT_READ:
            poller_events |= self._EVENT_READ
        if events & EVENT_WRITE:
            poller_events |= self._EVENT_WRITE
        try:
>           self._selector.register(key.fd, poller_events)
E           PermissionError: [Errno 1] Operation not permitted
/usr/local/lib/python3.10/selectors.py:360: PermissionError
```

bblommers:
Partially reverts #470 

Fixes #521 
Fixes #522 

@marcinsulikowski @boxed As the people that were originally involved with #470  - do you have any problems with this change, to gatekeep this behaviour behind a flag?

The linked issues indicate that this introduced quite some unintended side effects. Hiding this behaviour behind a flag keeps the feature accessible, but also doesn't break everything for others.

I'm torn whether this constitutes a breaking change, or a bugfix. Considering the amount of stuff it seems to break, I may just release it as a new minor version.

### DIFF
--- a/freezegun/api.py
+++ b/freezegun/api.py
@@ -544,7 +544,7 @@ class StepTickTimeFactory:
 
 class _freeze_time:
 
-    def __init__(self, time_to_freeze_str, tz_offset, ignore, tick, as_arg, as_kwarg, auto_tick_seconds):
+    def __init__(self, time_to_freeze_str, tz_offset, ignore, tick, as_arg, as_kwarg, auto_tick_seconds, real_asyncio):
         self.time_to_freeze = _parse_time_to_freeze(time_to_freeze_str)
         self.tz_offset = _parse_tz_offset(tz_offset)
         self.ignore = tuple(ignore)
@@ -554,6 +554,7 @@ class _freeze_time:
         self.modules_at_start = set()
         self.as_arg = as_arg
         self.as_kwarg = as_kwarg
+        self.real_asyncio = real_asyncio
 
     def __call__(self, func):
         if inspect.isclass(func):
@@ -727,20 +728,21 @@ class _freeze_time:
                         setattr(module, attribute_name, fake)
                         add_change((module, attribute_name, attribute_value))
 
-        # To avoid breaking `asyncio.sleep()`, let asyncio event loops see real
-        # monotonic time even though we've just frozen `time.monotonic()` which
-        # is normally used there. If we didn't do this, `await asyncio.sleep()`
-        # would be hanging forever breaking many tests that use `freeze_time`.
-        #
-        # Note that we cannot statically tell the class of asyncio event loops
-        # because it is not officially documented and can actually be changed
-        # at run time using `asyncio.set_event_loop_policy`. That's why we check
-        # the type by creating a loop here and destroying it immediately.
-        event_loop = asyncio.new_event_loop()
-        event_loop.close()
-        EventLoopClass = type(event_loop)
-        add_change((EventLoopClass, "time", EventLoopClass.time))
-        EventLoopClass.time = lambda self: real_monotonic()
+        if self.real_asyncio:
+            # To avoid breaking `asyncio.sleep()`, let asyncio event loops see real
+            # monotonic time even though we've just frozen `time.monotonic()` which
+            # is normally used there. If we didn't do this, `await asyncio.sleep()`
+            # would be hanging forever breaking many tests that use `freeze_time`.
+            #
+            # Note that we cannot statically tell the class of asyncio event loops
+            # because it is not officially documented and can actually be changed
+            # at run time using `asyncio.set_event_loop_policy`. That's why we check
+            # the type by creating a loop here and destroying it immediately.
+            event_loop = asyncio.new_event_loop()
+            event_loop.close()
+            EventLoopClass = type(event_loop)
+            add_change((EventLoopClass, "time", EventLoopClass.time))
+            EventLoopClass.time = lambda self: real_monotonic()
 
         return freeze_factory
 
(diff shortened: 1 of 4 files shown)


spt29:
I would very much appreciate this change: We had literally dozens of failures in our test runs due to the change in behaviour, so I would consider this change here as a bug fix. If one really needs the asyncio stuff, an explicit opt-in looks right. Otherwise pinning to 1.2.2 seems to be the only sane option for us at the moment. :disappointed:

PR #526, opened 2023-12-18
Fix local timezone and timezone related conversions

priezz:


bblommers (changes requested):
Hi @priezz, thaank you for the PR! At a first glance, this seems mostly sensible - but I would have to do a second (/third) deep-dive, as there seem to be quite a few functional changes.

Can you extend the tests to account for all changes? That makes it easier to verify that everything works as expected.

priezz:
> Hi @priezz, thaank you for the PR! At a first glance, this seems mostly sensible - but I would have to do a second (/third) deep-dive, as there seem to be quite a few functional changes.
> 
> Can you extend the tests to account for all changes? That makes it easier to verify that everything works as expected.

Sure, if after the deep dive you will confirm that the changes make sense, I will add missing/update existing tests.

bblommers (changes requested):

PR #527, opened 2024-01-01
rename excluded package "Queue" to match its python 3 name

yonil:
On python 3 `Queue` was renamed to `queue`. As a result it seems this is no longer excluding the queue package.


bblommers:
One of our tests is failing with this change: 

https://github.com/spulec/freezegun/actions/runs/7378273067/job/20138768211?pr=527

yonil:
@bblommers fixed the test. Should be good now.

bblommers (approved):
Nice one - thanks for the fix @yonil!

Issue #499, opened 2023-04-03
`extend_ignore_list` should add ignore package only once


lleontop:
Hi,

when configuring freezegun to extend the ignore list with `extend_ignore_list` there isn't any check if the same package(s) already exist in the list with the result that multiple invocations with the same parameter(s) cause the list to include multiple duplicates.

example code:
```python
>>> freezegun.configure(extend_ignore_list=['pymongo'])
>>> freezegun.configure(extend_ignore_list=['pymongo'])
>>> freezegun.config.settings.default_ignore_list
['nose.plugins', 'six.moves', 'django.utils.six.moves', ..., 'gi', 'pymongo', 'pymongo']
```

I think either checking for existence or using a set for a list could solve this. I can create a PR for that if you agree!

Thanks!

bblommers:
That makes sense @lleontop - PR's are welcome!

PR #530, opened 2024-01-21
Ignore duplicate modules in ignore list

lleontop:
Ignore duplicate modules passed in ignore_list. This could also be done with calling `set` on the final ignore list but sets are unordered which in terms of functionality I reckon should be ok but it was a pain in tests, so opted for a this `dict.fromkeys` way since supported python versions are >=3.7

fixes: #499


bblommers (approved):
LGTM - thanks for the PR @lleontop!

Issue #350, opened 2020-05-22
move_to not supported when ticks enabled


thehesiod:
it seems like only `StepTickTimeFactory` and `FrozenDateTimeFactory` have this method enabled but not `TickingDateTimeFactory`

spulec:
Thanks for opening! I am marking this as an enhancement.

thehesiod:
is it really an enhancement? The docs say this should work.  Btw here's my hack around it:

```python
class WraptTickingDateTimeFactory(wrapt.ObjectProxy):
    # https://github.com/spulec/freezegun/issues/350
    def move_to(self, target_datetime):
        """Moves frozen date to the given ``target_datetime``"""
        self.__wrapped__.time_to_freeze = freezegun.api._parse_time_to_freeze(target_datetime)
        self.__wrapped__.start = freezegun.api.real_datetime.now()

    if isinstance(frozen_datetime, freezegun.api.TickingDateTimeFactory):
        frozen_datetime = WraptTickingDateTimeFactory(frozen_datetime)
```

lokhman:
What's the status of this? It's really frustrating, especially when used together with `pytest-freezegun` library, which does not allow to change the original factory of the `freezer` fixture on the fly.

boxed:
The status is there is no PR to fix it.

lokhman:
We need a PR! :)

ashishnitinpatil:
https://github.com/spulec/freezegun/pull/531
I'm not sure where to add tests for this. Also, is the reset of `start` really required?

ashishnitinpatil:
Created a PR for this - https://github.com/spulec/freezegun/pull/533
For the time being, to circumvent the issue, you could (if possible) simply add the argument `auto_tick_seconds=1` to your `freeze_time` call.

PR #531, opened 2024-02-11
++TickingDateTimeFactory.move_to

ashishnitinpatil:
added missing move_to, resolves https://github.com/spulec/freezegun/issues/350

### DIFF
--- a/freezegun/api.py
+++ b/freezegun/api.py
@@ -493,6 +493,10 @@ class TickingDateTimeFactory:
     def __call__(self):
         return self.time_to_freeze + (real_datetime.now() - self.start)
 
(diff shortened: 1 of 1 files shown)


bblommers (changes requested):
Hi @ashishnitinpatil, thank you for the PR!

The `move_to` should update the `start`-attribute as well, as that is used to calculate how much time has progressed.

Can we also get a test for this in `test_ticking.py`? That file contains a few other tests that you can use for inspiration.

ashishnitinpatil:
Thanks for the comments @bblommers. I created another (couldn't edit the branch) PR resolving the comment suggestions - https://github.com/spulec/freezegun/pull/533

PR #534, opened 2024-03-02
Only call uuid._load_system_functions() if it exists.

gpshead:
This private API has been a no-op for several releases and is going away in Python 3.13.


bblommers (approved):
Makes sense @gpshead, thanks for fixing this proactively!

notatallshaw:
Is there a realease planned soon for this? I was trying to update pip's CI tests to support Python 3.13, and this is currently a blocker.

bblommers:
1.5.0 has just been released @notatallshaw

PR #538, opened 2024-03-20
Fix time_ns() freeze rounding error

romuald:
The previous behavior of time_ns() mock did round the value of the frozen date to the seconds, excluding the microseconds if specified

This is mostly problematic when freeze_time() is used with the uuid1() method: multiple "frozen" dates within the same second would all generate the same uuid


tests/test_datetimes.py
```diff
@@ -759,14 +759,25 @@ def test_time_ns():
     utc_time = local_time - datetime.timedelta(seconds=time.timezone)
     expected_timestamp = time.mktime(utc_time.timetuple())
 
-    freezer.start()
-    assert time.time() == expected_timestamp
-    assert time.time_ns() == expected_timestamp * 1e9
-    freezer.stop()
+    with freezer:
```
romuald, line 762: This is totally unrelated: this test will crash pytest if the time_ns() function raises an error

kingbuzzman: I'm not sure I understand your comment. On an error, wouldn't the `__exit__` code run, and thus NOT cause any other test side effects? I'm failing to see an issue, could you please elaborate?

bblommers (approved):
Makes sense - thanks for fixing this @romuald!

tests/test_datetimes.py
romuald: The issue was present _before_ the change. My comment should have said "did" crash and not "will" crash 😑

PR #539, opened 2024-03-21
Add tick() method to TickingDateTimeFactory

vicfergar:
Adds this missing method to `TickingDateTimeFactory`


bblommers (approved):
LGTM! Thanks for adding this @vicfergar

PR #540, opened 2024-04-09
Allow `default_ignore_list` to be empty

encukou:
Hello,
Thank you for freezegun!

I wanted to disable the default ignore list, but I noticed that `configure(default_ignore_list=[])` is equivalent to `default_ignore_list=None`. I believe it should instead ignore nothing.

I added a `is not None`, and a test.
I added the same test for `test_extend_default_ignore_list`, but since None and [] are equivalent there, that doesn't need code changes.


kingbuzzman (approved):
LGTM, but i'd get rid of these comments, they add no value whatsoever. 2 cents.
```
    ['threading', 'tensorflow'],  # example from docs
    [],  # ignore nothing
```

```
    ['tensorflow'],  # example from docs
    [],  # ignore nothing extra
```

bblommers (approved):
Makes sense @encukou - thanks for fixing this!

PR #546, opened 2024-04-25
Fix `tick` delta type handling

robsdedude:
## `FrozenDateTimeFactory.tick`
The `delta` argument is capable of handling `float`s. In previous versions of freezgun, the `.pyi` type annotations were correctly reflecting that. For some reason, when moving the type annotations into the `.py` file, this information got lost.

Further, checking for `isinstance(delta, numbers.Real)` is probably not what was intended as `fraction.Fraction` is a subclass of `Real`, but will cause an error when passed into `datetime.timedelta(seconds=delta)`.

## `StepTickTimeFactory.tick`
The same issue with the type hint applies here.

Fruther, passing an integer/float `delta` would lead to that number being added to the frozen `datetime.datetime`, which is not a valid operation (`TypeError: unsupported operand type(s) for +: 'datetime.datetime' and 'int'`).

### DIFF
--- a/freezegun/api.py
+++ b/freezegun/api.py
@@ -510,9 +510,10 @@ class TickingDateTimeFactory:
         return self.time_to_freeze + (real_datetime.now() - self.start)
 
     def tick(self, delta: Union[datetime.timedelta, float]=datetime.timedelta(seconds=1)) -> datetime.datetime:
-        if isinstance(delta, numbers.Real):
-            # noinspection PyTypeChecker
-            self.move_to(self.time_to_freeze + datetime.timedelta(seconds=delta))
+        if isinstance(delta, numbers.Integral):
+            self.move_to(self.time_to_freeze + datetime.timedelta(seconds=int(delta)))
+        elif isinstance(delta, numbers.Real):
+            self.move_to(self.time_to_freeze + datetime.timedelta(seconds=float(delta)))
         else:
             self.move_to(self.time_to_freeze + delta)  # type: ignore
         return self.time_to_freeze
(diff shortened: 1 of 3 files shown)


freezegun/api.py
```diff
@@ -531,8 +531,8 @@ def __init__(self, time_to_freeze: datetime.datetime):
     def __call__(self) -> datetime.datetime:
         return self.time_to_freeze
 
-    def tick(self, delta: Union[datetime.timedelta, int]=datetime.timedelta(seconds=1)) -> datetime.datetime:
-        if isinstance(delta, numbers.Real):
+    def tick(self, delta: Union[datetime.timedelta, float]=datetime.timedelta(seconds=1)) -> datetime.datetime:
+        if isinstance(delta, float):
```
bblommers, line 535: It looks like this change causes the tests to fail.

If passing fractions causes this bit to fail, I think we should tackle that separately - this PR can just focus on the type fix.

freezegun/api.py
```diff
@@ -557,9 +557,11 @@ def __call__(self) -> datetime.datetime:
         self.tick()
         return return_time
 
-    def tick(self, delta: Union[datetime.timedelta, int, None]=None) -> datetime.datetime:
+    def tick(self, delta: Union[datetime.timedelta, float, None]=None) -> datetime.datetime:
         if not delta:
             delta = datetime.timedelta(seconds=self.step_width)
+        elif isinstance(delta, float):
```
bblommers, line 563: Would you mind adding a test case for this?

bblommers (changes requested):
Hi @robsdedude, thanks for the PR! Please see my comments

freezegun/api.py
robsdedude: Sorry, my bad, it should be `isinstance(delta, (int, float))`. I got confused by `mypy` treating `int` as a subtype of `float`, but `isinstance` (correctly) doesn't.

As you suggested, I will split this off into another PR though. I theory, this could break people that managed to pass in some type that `isinstance` of `numbers.Real` *and* works with `datetime.timedelta(secods=x)`. Though, I couldn't come up with any such type other than `int` and `float`.

robsdedude:
I came up with what I think is a way to keep the runtime functionality as is, and get the type hints to be somewhat :tm: accurate.

bblommers (approved):
Looks great, thank you for fixing this and for adding the tests @robsdedude!

PR #550, opened 2024-06-28
Admin: Support Python 3.13

bblommers:


tests/test_datetimes.py
```diff
@@ -534,17 +534,17 @@ def test_method_decorator_works_on_unittest(self) -> None:
...
     def test_method_decorator_works_on_unittest_kwarg_hello(self, **kwargs: Any) -> None:
         self.assertEqual(datetime.date(2013, 4, 9), datetime.date.today())
-        self.assertEqual(datetime.date(2013, 4, 9), kwargs.get('hello').time_to_freeze.today())  # type: ignore
+        assert kwargs.get("hello").time_to_freeze.today().strftime('%Y-%m-%d') == "2013-04-09"  # type: ignore
 
     @freeze_time(lambda: datetime.date(year=2013, month=4, day=9), as_kwarg='frozen_time')
     def test_method_decorator_works_on_unittest_kwarg_frozen_time_with_func(self, frozen_time: Any) -> None:
         self.assertEqual(datetime.date(2013, 4, 9), datetime.date.today())
-        self.assertEqual(datetime.date(2013, 4, 9), frozen_time.time_to_freeze.today())
+        assert frozen_time.time_to_freeze.today().strftime('%Y-%m-%d') == "2013-04-09"
```
jayaddison, line 547: @bblommers were these test/assertion changes applied to make the test suite more tolerant to different timezones?

I'm finding that they tend to pass for negative timezone offsets (all the way up to `TZ='UTC-23:59'`, but for any positive timezone offset, they fail.

Ref: Debian bug https://bugs.debian.org/1106274

bblommers: @jayaddison They were changed because the original asserts started failing in Python 3.13. There is undoubtedly due to a bug somewhere, I just haven't the time or energy to look into it - and changing the assertions was the easiest way to make it work.

jayaddison: @bblommers Ok, gotcha - from testing that (which I probably should've done before commenting): comparison between a `datetime.date` and `datetime.datetime` no longer implicitly converts the latter to a `date` for comparison purposes in Py3.13 - so that makes sense.

I'll try to investigate what's going on with the timezone-related failure.

jayaddison: Removing the calls to `.today()` -- a classmethod -- resolves the problem, allowing the tests to pass for all of UTC+1/UTC-1/UTC when configured by the `TZ` environment variable.

But I haven't yet determined where in the code (perhaps even in cPython's local time adjustment logic?) the subtraction of an hour from `2013-04-09` (date) occurs when the `TZ` offset is positive -- e.g. Westwards, `TZ='UTC+1' pytest tests/test_datetimes.py` for example).

jayaddison: Ok, I have a suggestion that I think would maintain the previous intent of these tests, and is backwards compatible with pre-py3.13 while also fixing the bug.

The short story is that `time_to_freeze.today()` calls are problematic, because `today()` returns a `datetime`, that in Py3.13 fails equality tests against `date` objects.

It's also worth noting that `time_to_freeze` is a regular non-fake `datetime` itself; in practice `freezegun` does appear to encapsulate the class-level `today()` call made using it -- but some of the Python-level TZ adjustment occurs before a result is received, and this isn't what we want.

Reading the logic of the test itself, I think that what was intended was to compare `the date that we are frozen to` against a native `datetime.date` object.  That'd be a more appropriate case to use the [_instance_-level `.date()` method](https://docs.python.org/3/library/datetime.html#datetime.datetime.date) instead of the _class_-level `.today()`.

I'll offer a changeset with that as a PR, and will update the `README` example too.

PR #572, opened 2025-05-23
Tests: fixup: replace 'datetime.today' classmethod calls with 'datetime.date' method

jayaddison:
This provides an alternative resolution for a Python3.13 upgrade compatibility issue, by ensuring that both sides of the comparison are `datetime.date` objects.  Previously we had been calling `datetime.today()` that would return a `datetime.datetime` instance as the r-value.

This should also allow the test suite to pass regardless of the `TZ` (timezone) configuration of the test runner/system.

Ref: #550


jayaddison:
You're very welcome - thank you @bblommers!

PR #574, opened 2025-06-04
Restore class-decorator overload annotation for freeze_time

sfreilich:
This was removed in f9235ae, apparently inadvertently.


sfreilich:
Is this just about avoiding a warning on MyPy for Python 3.13? The problem is, this breaks Pytype, it no longer knows that the decorated class has the same type.

bblommers (approved):
Thanks for the PR @sfreilich - this does make sense. Thanks for adding an explanation as well, that should avoid us from making the same mistake again :slightly_smiling_face:

Issue #578, opened 2025-07-16
freezegun `1.5.3` breaks pytest yield fixtures


admirito-ultramarin:
The following code worked fine with freezegun `1.5.2`:

```python
from freezegun import freeze_time
import pytest

@freeze_time("2023-01-01")
class TestFoo:
    @pytest.fixture
    def myfoo(self):
        yield "foo"

    def test_foo(self, myfoo):
        assert myfoo == "foo"
```

However, in `1.5.3`, it produces this exception:

`AssertionError: assert <generator object TestFoo.myfoo at 0x104cc8b80> == 'foo'`

bblommers:
This is now fixed in `1.5.4`/`1.5.5`.

PR #579, opened 2025-07-28
fix: preserve functionality in pytest yield fixtures

zsh8:
Addresses issue #578.

Previously, pytest yield fixtures lost their behavior when decorated, causing test failures. This fix ensures they maintain generator capabilities, yielding as intended.


bblommers (approved):
LGTM - thanks for fixing this @zsh8!

PR #588, opened 2025-11-06
Implement datetime.datetime.combine

NHellFire:
datetime.datetime.combine() has type checking, so we need to implement this ourselves:
```
TypeError: combine() argument 2 must be datetime.time, not FakeDatetime
```


NHellFire:
Never mind, I accidentally passed a datetime.datetime, not datetime.time 😅

PR #592, opened 2026-02-13
Address sporadic "dictionary changed size during iteration" RuntimeError

### LINKED ISSUE #569: [Py3.11] RuntimeError: dictionary changed size during iteration
Hi team,

I'm getting this error intermittently:
```python
  File <SNIP>
    with freezegun.freeze_time(lambda: datetime(today.year,
  File " <SNIP>/python3.11/site-packages/freezegun/api.py", line 633, in __enter__
    return self.start()
           ^^^^^^^^^^^^
  File "<SNIP>/python3.11/site-packages/freezegun/api.py", line 714, in start
    for mod_name, module in list(sys.modules.items()):
                            ^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: dictionary changed size during iteration
```
Unfortunately, I don't have a reproducer yet. Would something like this work [here](https://github.com/spulec/freezegun/blob/master/freezegun/api.py#L773)?

```python
mod_names = list(sys.modules.keys())
for mod_name in mod_names:
    module =  sys.modules[mod_name]

    if mod_name is None or module is None or mod_name == __name__:
        continue
    elif mod_name.startswith(self.ignore) or mod_name.endswith('.six.moves'):
        continue
    elif (not hasattr(module, "__name__") or module.__name__ in ('datetime', 'time')):
        continue

    module_attrs = _get_cached_module_attributes(module)
    for attribute_name, attribute_value in module_attrs:
        fake = fakes.get(id(attribute_value))
        if fake:
            setattr(module, attribute_name, fake)
            add_change((module, attribute_name, attribute_value))

```

Please let me know if I'm missing something.

ento:
Closes #569

I started seeing this error in a test suite running on Python 3.10 with no discernible pattern and couldn't figure out a reliable reproducer like other commenters in the issue above.

[The official documentation](https://docs.python.org/3/library/sys.html#sys.modules) recommends doing `sys.modules.copy()` when iterating over it, so I figured it's both safe and worth trying.


chris-griffin:
I can confirm that this fix addresses the RuntimeError across a test of 300k iterations against freezegun 1.5.5 running python 3.11 in our code.

Before patch:
322 RuntimeError  (0.1073%)

After patch:
0 RuntimeError  (0.0000%)

